PR #2, opened 2016-03-13
Reconfigure projects and targets; update dependencies.

hiltonc:


hiltonc:
I learned how to do things better with LDSContent, this mostly just adjusts things to match. For example, after this change we won't need to `pod install` after adding a file to the LDSAnnotations framework.

stephanheilner:
:+1:

PR #3, opened 2016-04-13
Sync Annotations

stephanheilner:
Missing the UI and tests still, but sync works.


LDSAnnotations/Annotation.swift
```diff
@@ -0,0 +1,183 @@
...
+        self.lastModified = lastModified
+        self.noteID = noteID
+        self.bookmarkID = bookmarkID
+        self.source = source
+        self.device = device
+    }
+    
+    init?(jsonObject: [String: AnyObject]) {
+        guard let uniqueID = jsonObject["@id"] as? String,
+            type = jsonObject["@type"] as? String, annotationType = AnnotationType(rawValue: type),
```
nickmshelley: Why are some on newlines and some combined? Can we either do one per line or all on one line?

LDSAnnotations/Annotation.swift
```diff
@@ -0,0 +1,183 @@
...
+        }
+        
+        self.id = nil
+        self.uniqueID = uniqueID
+        self.type = annotationType
+        self.lastModified = lastModified
+        self.docID = docID
+        self.docVersion = docVersion
+        
+        if let rawCreated = jsonObject["created"] as? String, created = NSDate.parseFormattedISO8601(rawCreated) {
```
nickmshelley: `self.created = (jsonObject["created"] as? String).flatMap { NSDate.parseFormattedISO8601($0) }` (if you want).

LDSAnnotations/Annotation.swift
```diff
@@ -0,0 +1,183 @@
...
+            self.created = created
+        } else {
+            self.created = nil
+        }
+        
+        if let iso639_3Code = jsonObject["@locale"] as? String {
```
nickmshelley: `self. iso639_3Code = jsonObject["@locale"] as? String ?? "eng"`

LDSAnnotations/Annotation.swift
```diff
@@ -0,0 +1,183 @@
...
+            self.iso639_3Code = iso639_3Code
+        } else {
+            self.iso639_3Code = "eng"
+        }
+
+        if let rawStatus = jsonObject["@status"] as? String, status = AnnotationStatus(rawValue: rawStatus) {
```
nickmshelley: `self.status = jsonObject["@status"] as? String ?? .Active`

LDSAnnotations/Annotation.swift
```diff
@@ -0,0 +1,183 @@
...
+            self.status = status
+        } else {
+            self.status = .Active
+        }
+        
+        if let source = jsonObject["source"] as? String {
```
nickmshelley: `self.source = jsonObject["source"] as? String`

LDSAnnotations/Annotation.swift
```diff
@@ -0,0 +1,183 @@
...
+            self.source = source
+        } else {
+            self.source = nil
+        }
+        
+        if let device = jsonObject["device"] as? String {
```
nickmshelley: `self.device = jsonObject["device"] as? String`

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -0,0 +1,379 @@
...
+    /// Trashes the `annotations`; throws if any of the `annotations` are not active or trashed.
+    public func trashAnnotations(annotations: [Annotation]) throws {
+        try inTransaction {
+            let lastModified = NSDate()
+            
+            var ids = [Int64]()
+            
+            // Fetch the current status of the annotations
+            for annotation in self.allAnnotations(ids: annotations.flatMap { $0.id }) {
+                if annotation.status != .Active && annotation.status != .Trashed {
```
nickmshelley: This trashes the annotation but fails if it's not already trashed...

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -0,0 +1,379 @@
...
+                    throw Error.errorWithCode(.Unknown, failureReason: "Attempted to trash a annotation that is not active or trashed.")
+                }
+                
+                if let id = annotation.id {
```
nickmshelley: `annotation.id` is not optional so it doesn't make sense to `if let` here.

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -0,0 +1,379 @@
...
+                    try self.db.run(AnnotationTable.table.filter(AnnotationTable.id == id).update(
+                        AnnotationTable.status <- .Trashed,
+                        AnnotationTable.lastModified <- lastModified
+                    ))
+                    
+                    ids.append(id)
```
nickmshelley: Since all ids are guaranteed to be there, you can get rid of this temporary array.

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -0,0 +1,379 @@
...
+    /// Deletes the `annotations`; throws if any of the `annotations` are not trashed or deleted.
+    public func deleteAnnotations(annotations: [Annotation]) throws {
+        try inTransaction {
+            let lastModified = NSDate()
+            
+            var ids = [Int64]()
+            
+            // Fetch the current status of the annotations
+            for annotation in self.allAnnotations(ids: annotations.flatMap { $0.id }) {
+                if annotation.status != .Trashed && annotation.status != .Deleted {
```
nickmshelley: Doesn't make sense to error if it's not deleted.

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -0,0 +1,379 @@
...
+                    throw Error.errorWithCode(.Unknown, failureReason: "Attempted to delete a annotation that is not trashed or deleted.")
+                }
+                
+                if let id = annotation.id {
```
nickmshelley: `id` isn't optional.

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -0,0 +1,379 @@
...
+                    try self.db.run(AnnotationTable.table.filter(AnnotationTable.id == id).update(
+                        AnnotationTable.status <- .Deleted,
+                        AnnotationTable.lastModified <- lastModified
+                    ))
+                    
+                    ids.append(id)
```
nickmshelley: Get rid of this.

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -0,0 +1,379 @@
...
+    /// Restores the `annotations` to active.
+    public func restoreAnnotations(annotations: [Annotation]) throws {
+        try inTransaction {
+            let lastModified = NSDate()
+            
+            var ids = [Int64]()
+            
+            // Fetch the current status of the annotations
+            for annotation in self.allAnnotations(ids: annotations.flatMap { $0.id }) {
+                if let id = annotation.id {
```
nickmshelley: `id` isn't optional.

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -0,0 +1,379 @@
...
+    
+    /// Restores the `annotations` to active.
+    public func restoreAnnotations(annotations: [Annotation]) throws {
+        try inTransaction {
+            let lastModified = NSDate()
+            
+            var ids = [Int64]()
+            
+            // Fetch the current status of the annotations
+            for annotation in self.allAnnotations(ids: annotations.flatMap { $0.id }) {
```
nickmshelley: `id` isn't optional so just `map`.

LDSAnnotations/Annotation.swift
```diff
@@ -0,0 +1,183 @@
...
+// THE SOFTWARE.
+//
+
+import Foundation
+
+/// An annotation.
+public struct Annotation {
+    
+    /// Local ID.
+    public internal(set) var id: Int64?
```
nickmshelley: This shouldn't be optional.

nickmshelley: Stephan explained why this is optional. I'll go delete my other comments.

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -0,0 +1,379 @@
...
+            }
+            
+            self.notifyModifiedAnnotationsWithIDs(ids)
+        }
+    }
+    
+    func allAnnotations(ids ids: [Int64]? = nil, lastModifiedAfter: NSDate? = nil, lastModifiedOnOrBefore: NSDate? = nil) -> [Annotation] {
+        do {
+            var query = AnnotationTable.table
+            if let ids = ids {
```
nickmshelley: You can use `flatMap` to get rid of these `if let`s if you want: `query = ids.flatMap { query.filter($0.contains(AnnotationTable.id)) }`

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -0,0 +1,379 @@
...
+            }
+            return try db.prepare(query).map { AnnotationTable.fromRow($0) }
+        } catch {
+            return []
+        }
+    }
+    
+    func deletedAnnotations(lastModifiedOnOrBefore lastModifiedOnOrBefore: NSDate) -> [Annotation] {
+        do {
+            return try db.prepare(AnnotationTable.table.filter(AnnotationTable.lastModified <= lastModifiedOnOrBefore && AnnotationTable.status == .Deleted)).map { AnnotationTable.fromRow($0) }
```
nickmshelley: Seems like it would be more efficient to put the status condition first, but maybe not (and it probably doesn't matter).

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -0,0 +1,379 @@
...
+        } catch {
+            return []
+        }
+    }
+    
+    func annotationWithUniqueID(uniqueID: String) -> Annotation? {
+        return db.pluck(AnnotationTable.table.filter(AnnotationTable.uniqueID == uniqueID)).map { AnnotationTable.fromRow($0) }
```
nickmshelley: This doesn't seem like it should work since `pluck` returns a single row (I think).

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -0,0 +1,379 @@
...
+            return []
+        }
+    }
+    
+    func deleteAnnotationWithID(id: Int64) {
+        do {
+            try db.run(AnnotationTable.table.filter(AnnotationTable.id == id).delete())
+        } catch {}
+        
+        notifySyncModifiedAnnotationsWithIDs([id])
```
nickmshelley: You should probably put this inside the `do` since if it fails nothing will have been modified.

LDSAnnotations/AnnotationStore+AnnotationNotebook.swift
```diff
@@ -0,0 +1,80 @@
...
+    func createAnnotationNotebookTable() throws {
+        try db.run(AnnotationNotebookTable.table.create(ifNotExists: true) { builder in
+            builder.column(AnnotationNotebookTable.annotationID, references: AnnotationTable.table, AnnotationTable.id)
+            builder.column(AnnotationNotebookTable.notebookID, references: NotebookTable.table, NotebookTable.id)
+            builder.column(AnnotationNotebookTable.displayOrder)
+            builder.primaryKey(AnnotationNotebookTable.annotationID, AnnotationNotebookTable.notebookID)
+        })
+    }
+    
+    /// Adds a new folder with `name`.
```
nickmshelley: There's no `name` in the function.

LDSAnnotations/AnnotationStore+AnnotationNotebook.swift
```diff
@@ -0,0 +1,80 @@
...
+                AnnotationNotebookTable.displayOrder <- displayOrder
+            ))
+            
+            return AnnotationNotebook(annotationID: annotationID, notebookID: notebookID, displayOrder: displayOrder)
+        } catch {
+            return nil
+        }
+    }
+    
+    func deleteAnnotationNotebookWithID(annotationID: Int64, notebookID: Int64) {
```
nickmshelley: Remove the `WithID` part of the name since that's old style.

LDSAnnotations/AnnotationStore+AnnotationTag.swift
```diff
@@ -0,0 +1,77 @@
...
+    
+    func createAnnotationTagTable() throws {
+        try db.run(AnnotationTagTable.table.create(ifNotExists: true) { builder in
+            builder.column(AnnotationTagTable.annotationID, references: AnnotationTable.table, AnnotationTable.id)
+            builder.column(AnnotationTagTable.tagID, references: TagTable.table, TagTable.id)
+            builder.primaryKey(AnnotationTagTable.annotationID, AnnotationTagTable.tagID)
+        })
+    }
+    
+    /// Adds a new tag with `name`.
```
nickmshelley: There's no `name` in the function.

LDSAnnotations/AnnotationStore+Bookmark.swift
```diff
@@ -0,0 +1,98 @@
...
+        guard bookmark.annotationID != 0 else {
+            throw Error.errorWithCode(.Unknown, failureReason: "Cannot add a bookmark without an annotation ID.")
+        }
+        
+        do {
+            if let id = bookmark.id {
+                try db.run(BookmarkTable.table.filter(BookmarkTable.id == id).update(
+                    BookmarkTable.name <- bookmark.name,
+                    BookmarkTable.paragraphAID <- bookmark.paragraphAID,
+                    BookmarkTable.displayOrder <- bookmark.displayOrder
```
nickmshelley: Why aren't you updating the annotationID here?

LDSAnnotations/AnnotationStore+Note.swift
```diff
@@ -0,0 +1,92 @@
...
+    public func addOrUpdateNote(note: Note) throws -> Note? {
+        guard !note.content.isEmpty && note.annotationID != 0 else {
+            throw Error.errorWithCode(.Unknown, failureReason: "Cannot add a note without content and an annotation ID.")
+        }
+        
+        do {
+            if let id = note.id {
+                try db.run(NoteTable.table.filter(NoteTable.id == id).update(
+                    NoteTable.title <- note.title,
+                    NoteTable.content <- note.content
```
nickmshelley: Why aren't you updating the annotationID here? Will it never change?

LDSAnnotations/AnnotationStore+Notebook.swift
```diff
@@ -0,0 +1,334 @@
...
+    /// Trashes the `notebooks`; throws if any of the `notebooks` are not active or trashed.
+    public func trashNotebooks(notebooks: [Notebook]) throws {
+        try inTransaction {
+            let lastModified = NSDate()
+            
+            var ids = [Int64]()
+            
+            // Fetch the current status of the notebooks
+            for notebook in self.allNotebooks(ids: notebooks.flatMap { $0.id }) {
+                if notebook.status != .Active && notebook.status != .Trashed {
```
nickmshelley: Why does it fail if it's not trashed?

LDSAnnotations/AnnotationStore+Notebook.swift
```diff
@@ -0,0 +1,334 @@
...
+    /// Deletes the `notebooks`; throws if any of the `notebooks` are not trashed or deleted.
+    public func deleteNotebooks(notebooks: [Notebook]) throws {
+        try inTransaction {
+            let lastModified = NSDate()
+            
+            var ids = [Int64]()
+            
+            // Fetch the current status of the notebooks
+            for notebook in self.allNotebooks(ids: notebooks.flatMap { $0.id }) {
+                if notebook.status != .Trashed && notebook.status != .Deleted {
```
nickmshelley: Why does it fail if it's not already deleted?

LDSAnnotations/AnnotationStore+Notebook.swift
```diff
@@ -0,0 +1,334 @@
...
+            }
+            
+            self.notifyModifiedNotebooksWithIDs(ids)
+        }
+    }
+    
+    func allNotebooks(ids ids: [Int64]? = nil, lastModifiedAfter: NSDate? = nil, lastModifiedOnOrBefore: NSDate? = nil) -> [Notebook] {
+        do {
+            var query = NotebookTable.table
+            if let ids = ids {
```
nickmshelley: You can change these `if let`s to `flatMap`s if you want.

LDSAnnotations/AnnotationStore+Notebook.swift
```diff
@@ -0,0 +1,334 @@
...
+            return nil
+        }
+    }
+    
+    func deleteNotebookWithID(id: Int64) {
+        do {
+            try db.run(NotebookTable.table.filter(NotebookTable.id == id).delete())
+        } catch {}
+        
+        notifySyncModifiedNotebooksWithIDs([id])
```
nickmshelley: This should go in the `do` since if it fails nothing will have been updated.

LDSAnnotations/Bookmark.swift
```diff
@@ -0,0 +1,90 @@
...
+    }
+    
+    init?(jsonObject: [String: AnyObject], annotationID: Int64) {
+        guard let name = jsonObject["name"] as? String else { return nil }
+        
+        self.id = nil
+        self.name = name
+        self.annotationID = annotationID
+        
+        if let offset = jsonObject["@offset"] as? Int {
```
nickmshelley: `self.offset = jsonObject["@offset"] as? Int`

LDSAnnotations/Bookmark.swift
```diff
@@ -0,0 +1,90 @@
...
+            self.offset = offset
+        } else {
+            self.offset = nil
+        }
+        
+        if let displayOrder = jsonObject["sort"] as? Int {
```
nickmshelley: Similar to above.

LDSAnnotations/Bookmark.swift
```diff
@@ -0,0 +1,90 @@
...
+            self.displayOrder = displayOrder
+        } else {
+            self.displayOrder = nil
+        }
+    }
+    
+    func jsonObject() -> [String: AnyObject] {
+        var result: [String: AnyObject] = ["@offset": offset ?? -1]
```
nickmshelley: It seems like it would be better to not include the `@offset` key instead of making it `-1`.

LDSAnnotations/Highlight.swift
```diff
@@ -0,0 +1,93 @@
...
+        }
+        
+        self.id = nil
+        self.paragraphAID = paragraphAID
+        self.offsetStart = offsetStart
+        self.offsetEnd = offsetEnd
+        self.colorName = colorName
+        self.annotationID = annotationID
+        
+        if let rawStyle = jsonObject["@style"] as? String, style = HighlightStyle(rawValue: rawStyle) {
```
nickmshelley: `self.style = (jsonObject["@style"] as? String).flatMap { HighlightStyle(rawValue: $0) } ?? .Highlight`

LDSAnnotations/HighlightStyle.swift
```diff
@@ -0,0 +1,38 @@
...
+//
+
+import Foundation
+import SQLite
+
+/// The style of a highlight.
+public enum HighlightStyle: String {
+    
+    /// Plain block highlight style
+    case Highlight = ""
```
nickmshelley: Why the empty string?

LDSAnnotations/Link.swift
```diff
@@ -0,0 +1,74 @@
...
+        self.name = name
+        self.docID = docID
+        self.docVersion = docVersion
+        self.paragraphAIDs = paragraphAIDs.componentsSeparatedByString(",").map { $0.stringByTrimmingCharactersInSet(.whitespaceCharacterSet()) }
+        self.annotationID = annotationID
+    }
+    
+    func jsonObject() -> [String: AnyObject] {
+        return [
+            "$": name,
```
nickmshelley: What's with the key?

LDSAnnotations/Note.swift
```diff
@@ -0,0 +1,71 @@
...
+    }
+    
+    init?(jsonObject: [String: AnyObject], annotationID: Int64) {
+        guard let content = jsonObject["content"] as? String else { return nil }
+        
+        self.id = nil
+        self.content = content
+        self.annotationID = annotationID
+        
+        if let title = jsonObject["title"] as? String {
```
nickmshelley: `self.title = jsonObject["title"] as? String`

LDSAnnotations/Session.swift
```diff
@@ -79,14 +83,24 @@ public class Session: NSObject {
...
+    public func syncAnnotations(annotationStore annotationStore: AnnotationStore, token: SyncToken?, completion: (SyncNotebooksResult, SyncAnnotationsResult) -> Void) {
+        let syncNotebooksOperation = SyncNotebooksOperation(session: self, annotationStore: annotationStore, token: token) { syncNotebooksResult in
+            switch syncNotebooksResult {
+            case let .Success(localSyncDate: localSyncDate, serverSyncDate: serverSyncDate, notebookAnnotationIDs: notebookAnnotationIDs, uploadCount: _, downloadCount: _):
+                let syncAnnotationsOperation = SyncAnnotationsOperation(session: self, annotationStore: annotationStore, token: token, localSyncDate: localSyncDate, serverSyncDate: serverSyncDate, notebookAnnotationIDs: notebookAnnotationIDs) { syncAnnotationsResult in
+                    completion(syncNotebooksResult, syncAnnotationsResult)
+                }
+                self.operationQueue.addOperation(syncAnnotationsOperation)
+            case .Error(errors: _):
+                completion(syncNotebooksResult, SyncAnnotationsResult.Error(errors: []))
```
nickmshelley: Why are you sending an empty errors array instead of passing the actual errors through?

nickmshelley:
That was a doozy!

LDSAnnotations/Annotation.swift
```diff
@@ -0,0 +1,183 @@
...
+    /// ISO639-3 Language Code.
+    public var iso639_3Code: String
+
+    /// Annotation Type.
+    public var type: AnnotationType
+    
+    /// Whether the annotation is active, trashed, or deleted.
+    public internal(set) var status: AnnotationStatus
+    
+    /// When the annotation was last modified in local time.
```
hiltonc: /// When the annotation was created in local time.

LDSAnnotations/AnnotationIndex.swift
```diff
@@ -0,0 +1,32 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+import Foundation
+
+/// An annotation.
```
hiltonc: Can you expand a little in this comment? I don't know what an AnnotationIndex is (yet)

LDSAnnotations/AnnotationStore+Annotation.swift
hiltonc: It will only fail if it's not active and not trashed (i.e. deleted).

LDSAnnotations/AnnotationStore+Annotation.swift
hiltonc: Here, it only fail if it's active (not trashed and not deleted).

LDSAnnotations/AnnotationStore+Note.swift
```diff
@@ -0,0 +1,92 @@
...
+        } catch {
+            return nil
+        }
+    }
+    
+    public func noteWithID(id: Int64) -> Note? {
+        return db.pluck(NoteTable.table.filter(NoteTable.id == id)).map { NoteTable.fromRow($0) }
+    }
+    
+    public func deleteNoteWithID(id: Int64) {
```
hiltonc: Should this be internal?

LDSAnnotations/AnnotationStore.swift
```diff
@@ -66,14 +66,15 @@ public class AnnotationStore {
...
                 try db.transaction {
-                    try self.db.run(NotebookTable.table.create(ifNotExists: true) { builder in
-                        builder.column(NotebookTable.id, primaryKey: true)
-                        builder.column(NotebookTable.uniqueID)
-                        builder.column(NotebookTable.name)
-                        builder.column(NotebookTable.description)
-                        builder.column(NotebookTable.status)
-                        builder.column(NotebookTable.lastModified)
-                    })
+                    try self.createNotebookTable()
```
hiltonc: I like how you split each of these concerns into separate extension files. This file would have been very long otherwise.

LDSAnnotations/HighlightStyle.swift
```diff
@@ -0,0 +1,38 @@
...
+    
+    /// Underline style
+    case Underline = "red-underline"
```
hiltonc: That is awesome :)

LDSAnnotations/Link.swift
hiltonc: At least it's not "red-$"

LDSAnnotations/Notebook.swift
```diff
@@ -72,8 +72,8 @@ public struct Notebook {
         }
     }
     
-    func jsonObject() -> [String: AnyObject] {
-        var result = [
+    func jsonObject(annotationStore: AnnotationStore) -> [String: AnyObject] {
```
hiltonc: I understand why we need to pass in `annotationStore`, but it makes me wince. I can't think of a better alternative.

LDSAnnotations/AnnotationStore+Annotation.swift
nickmshelley: Ah that makes sense, thanks. I read that as or, not and.

Can we just change it to `if annotation.status == .Deleted`?

LDSAnnotations/AnnotationStore+Annotation.swift
nickmshelley: Can we change it to `if annotation.status == .Active`?

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -0,0 +1,379 @@
...
+    public func trashAnnotations(annotations: [Annotation]) throws {
+        try inTransaction {
+            let lastModified = NSDate()
+            
+            var ids = [Int64]()
+            
+            // Fetch the current status of the annotations
+            for annotation in self.allAnnotations(ids: annotations.flatMap { $0.id }) {
+                if annotation.status != .Active && annotation.status != .Trashed {
+                    throw Error.errorWithCode(.Unknown, failureReason: "Attempted to trash a annotation that is not active or trashed.")
```
nickmshelley: Oh, here's where I saw the "or." :) We'll need to update the error message too.

LDSAnnotations/SyncAnnotationsOperation.swift
```diff
@@ -0,0 +1,340 @@
...
+    var uploadTagCount = 0
+    
+    var downloadCount = 0
+    var downloadHighlightCount = 0
+    var downloadBookmarkCount = 0
+    var downloadNoteCount = 0
+    var downloadLinkCount = 0
+    var downloadTagCount = 0
+    
+    init(session: Session, annotationStore: AnnotationStore, token: SyncToken?, localSyncDate: NSDate?, serverSyncDate: NSDate?, notebookAnnotationIDs: [String: [String]]?, completion: (SyncAnnotationsResult) -> Void) {
```
hiltonc: You should be able to drop the localSyncDate and serverSyncDate parameters, and just use the token (which encapsulates them) instead. I was thinking that would be a useful thing to do so that users of this library don't need to worry how that all actually works, they just have a token they can easily persist.

LDSAnnotations/Session.swift
```diff
@@ -40,9 +40,12 @@ public class Session: NSObject {
     let clientPassword: String
     let authenticationURL: NSURL?
     let domain: String
+    let source: String
+    
+    var docVersionsForDocIDs: ((docIDs: [String]) -> ([String: Int]))?
```
hiltonc: Should this be required in the Session init?

hiltonc: Also, I don't think you could make a simpler API for this, I love it.

LDSAnnotations/SyncAnnotationsResult.swift
```diff
@@ -0,0 +1,50 @@
...
+import Foundation
+
+/// Result of syncing notebooks.
+public enum SyncAnnotationsResult {
+    
+    /// Result when sync is successful.
+    ///
+    /// The `token` should be used for the next sync.
+    case Success(token: SyncToken?,
+        localSyncDate: NSDate,
```
hiltonc: I think you can drop localSyncDate and serverSyncDate, and just have token.

hiltonc: Or, if you don't like having the token, you can remove it instead.

hiltonc:
Yeah, that was a lot to get through :). Awesome though!

LDSAnnotations/Bookmark.swift
stephanheilner: I would, but the docs say this:
```
offset (required attribute) - The word offset that marks the word for this bookmark. It is one based, so zero is not allowed. A value of -1 means the beginning of the paragraph.
Valid values: -1, 1-end
```

LDSAnnotations/HighlightStyle.swift
stephanheilner: That's just how the service works. D-U-M-B, I know. But basically empty string means highlight, 'red-underline' means underline.

LDSAnnotations/Link.swift
stephanheilner: yup, just service stupid-ness.

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -178,7 +178,7 @@ extension AnnotationStore {
             // Fetch the current status of the annotations
             for annotation in self.allAnnotations(ids: annotations.flatMap { $0.id }) {
                 if annotation.status != .Active && annotation.status != .Trashed {
```
nickmshelley: Can we make this `if annotation.status == .Deleted`?

LDSAnnotations/AnnotationStore+Annotation.swift
nickmshelley: Can we change the conditional to `if annotation.status == .Deleted`?

nickmshelley:
One minor thing then I think it's good to go. Nice work!

LDSAnnotations/SyncAnnotationsOperation.swift
```diff
@@ -0,0 +1,340 @@
...
+                    
+                    if let annotationID = databaseAnnotation?.id, annotationUniqueID = databaseAnnotation?.uniqueID {
+                        
+                        // Note
+                        if let note = annotation["note"] as? [String: AnyObject] {
+                            if let downloadedNote = Note(jsonObject: note, annotationID: annotationID) {
+                                try annotationStore.addOrUpdateNote(downloadedNote)
+                                downloadNoteCount += 1
+                            } else {
+                                NSLog("Failed to deserialize note: %@", note)
```
hiltonc: This should throw, not log.

Whenever things fail to deserialize, the entire operation needs to fail. This is because our sync mechanism is date-based, which means a sync has to be atomic.

We need to:

1. Make our sync code as rock solid as we can
2. Detect these errors and report them
3. Patch up the data in the back end (if the data itself is bad, perhaps due to a bad client), or improve the sync code in the next release (if the data is good, but reveals an edge case we hadn't considered)

hiltonc:
I pushed this branch to https://github.com/CrossWaterBridge/LDSAnnotations to trigger unit tests. Because they rely on encrypted data, only code pushed to this repo (as opposed to forks) triggers unit tests.

hiltonc:
🎤 💧 Sorry for piling a bunch on. @stephanheilner @nickmshelley plz to review.

hiltonc:
My bits are rotting!

LDSAnnotations/Bookmark.swift
```diff
@@ -40,25 +40,27 @@ public struct Bookmark {
     /// ID of annotation
     public var annotationID: Int64
     
-    /// The word offset that marks the word for this bookmark. It is one based, so zero is not allowed. A value of -1 means the beginning of the paragraph.
+    /// The word offset that marks the word for this bookmark. It is one based, so zero is not allowed. `nil` means the beginning of the paragraph.
```
nickmshelley: It seems like a zero offset should make sense and mean the beginning of a paragraph. Is this a legacy decision we just have to deal with?

LDSAnnotations/Bookmark.swift
```diff
@@ -78,3 +80,12 @@ public struct Bookmark {
     }
     
 }
+
+public func == (lhs: Bookmark, rhs: Bookmark) -> Bool {
+    return lhs.id == rhs.id
```
nickmshelley: Can't you just stop here?

LDSAnnotationsTests/AnnotationStoreTests.swift
```diff
@@ -0,0 +1,37 @@
...
+//
+
+import XCTest
+@testable import LDSAnnotations
+
+class AnnotationStoreTests: XCTestCase {
+    
+    func testAddNote() {
+        let annotationStore = AnnotationStore()!
+        let noteToAdd = Note(id: nil, title: nil, content: "", annotationID: 1)
```
nickmshelley: Why is this behavior desirable? It seems like it makes sense to treat empty notes as deletions as we do in GL.

LDSAnnotationsTests/SyncNotebooksTests.swift
```diff
@@ -139,18 +139,25 @@ class SyncNotebooksTests: XCTestCase {
...
     func sync(annotationStore: AnnotationStore, session: Session, inout token: SyncToken?, description: String, completion: (uploadCount: Int, downloadCount: Int) -> Void) {
         let expectation = expectationWithDescription(description)
-        session.syncNotebooks(annotationStore: annotationStore, token: token) { result in
-            switch result {
-            case let .Success(newToken, uploadCount, downloadCount):
+        session.syncAnnotations(annotationStore: annotationStore, token: token) { syncNotebooksResult, syncAnnotationsResult in
+            switch (syncNotebooksResult, syncAnnotationsResult) {
+            case let (
+                .Success(localSyncNotebooksDate: _, serverSyncNotebooksDate: _, notebookAnnotationIDs: _, uploadCount: notebookUploadCount, downloadCount: notebookDownloadCount),
+                .Success(token: newToken, uploadCount: _, uploadNoteCount: _, uploadBookmarkCount: _, uploadHighlightCount: _, uploadTagCount: _, uploadLinkCount: _, downloadCount: _, downloadNoteCount: _, downloadBookmarkCount: _, downloadHighlightCount: _, downloadTagCount: _, downloadLinkCount: _)
```
nickmshelley: I think it would simplify completion code if these were structs instead of tuples.

LDSAnnotations/SyncResult.swift
```diff
@@ -0,0 +1,50 @@
...
+
+import Foundation
+
+/// Result of syncing notebooks and annotations.
+public enum SyncResult {
+    
+    /// Result when sync is successful.
+    ///
+    /// The `token` should be used for the next sync.
+    case Success(token: SyncToken?,
```
nickmshelley: I'd definitely make this take a struct instead: `case Success(result: SyncResultStructWithBetterName)`

nickmshelley:
Some comments somewhere in some commits somewhere.

nickmshelley:
Oh apparently most of them are marked as outdated because they aren't the latest...

LDSAnnotations/Bookmark.swift
hiltonc: We could do that. I'm already converting -1 to `nil`. We could subtract 1 from what the server gives us and add 1 when sending it back.

LDSAnnotations/Bookmark.swift
hiltonc: I added this in particular so that I could use it in unit tests to verify that I got the right Bookmark. Just comparing `id` could give me false positives.

LDSAnnotationsTests/AnnotationStoreTests.swift
hiltonc: When I ran the code to download annotations on my account, I had notes with empty content. So I figured I'd adjust LDSAnnotations to allow them. I'm not sure what we should do if we want to treat them as deletions.

hiltonc: But if we do want to treat them as deletions, I can just update this unit test for the desired behavior.

LDSAnnotationsTests/SyncNotebooksTests.swift
hiltonc: Something like `.Success(token: newToken, counts: _)`?

hiltonc:
Fortunately I got a notification for each one :). Lots of questions in response. Lemmie know what you think.

LDSAnnotationsTests/SyncNotebooksTests.swift
nickmshelley: Yeah maybe. It just seems a bit unwieldily to have so many things, especially when they're ignored. Combining them somehow will let you use or ignore them as you will.

nickmshelley:
I'm good with this how it is, so 👍 unless @stephanheilner wants a go at it.

hiltonc:
@nickmshelley How's that?

nickmshelley:
Looks great!

stephanheilner:
👍 Love it, lets merge this bad boy!

PR #5, opened 2016-06-03
Added session controller, add AccountController to framework for app use

stephanheilner:


LDSAnnotations.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = "LDSAnnotations"
-  s.version      = "1.0.0"
+  s.version      = "1.1.0"
   s.summary      = "Swift client library for LDS annotation sync."
   s.author       = 'Hilton Campbell'
```
hiltonc: Put your name here too:

    s.authors = 'Stephan Heilner', 'Hilton Campbell'

LDSAnnotations/AccountController.swift
```diff
@@ -23,76 +23,69 @@
 import Foundation
 import Swiftification
 import Locksmith
-import LDSAnnotations
 
-class AccountController {
+public class AccountController {
     
-    static let sharedController = AccountController()
+    public static let sharedController = AccountController()
```
hiltonc: I think libraries shouldn't provide singletons. I'd rather see the app itself provide an extension if it wants a singleton.

LDSAnnotations/AccountController.swift
```diff
@@ -23,76 +23,69 @@
...
     
     private static let service = "LDSAccount"
+    private static let tokensKey = "tokens"
```
hiltonc: Declare this right before it's used, like usernamesKey

LDSAnnotations/AccountController.swift
```diff
@@ -23,76 +23,69 @@
...
-        
+    public func setSyncToken(token: SyncToken?, forUsername username: String) {
         var tokens = NSUserDefaults.standardUserDefaults().objectForKey(AccountController.tokensKey) as? [String: String] ?? [:]
         tokens[username] = token?.rawValue
         NSUserDefaults.standardUserDefaults().setObject(tokens, forKey: AccountController.tokensKey)
     }
     
+    public func annotationStoreForUsername(username: String) -> AnnotationStore? {
+        let storeName = String(format: "%@.sqlite", username)
+        return AnnotationStore(path: NSFileManager.privateDocumentsURL.URLByAppendingPathComponent(storeName).path)
```
hiltonc: I think "Private Documents" is a good practice (recommended by Apple in one of their tech notes), but it's not universally followed. Can you externalize that, maybe by having another arg to this method for the annotation stores directory, or having a delegate on AccountController for providing the annotation stores directory?

LDSAnnotations/Bookmark.swift
```diff
@@ -53,6 +53,11 @@ public struct Bookmark: Equatable {
...
     
     init?(jsonObject: [String: AnyObject], annotationID: Int64) {
+        if jsonObject["@pid"] == nil && jsonObject["uri"] != nil {
+            // Bookmark has a URI, but no @pid so its invalid
+            return nil
+        }
+        
         self.id = nil
         self.name = jsonObject["name"] as? String
         self.paragraphAID = jsonObject["@pid"] as? String
```
hiltonc: `paragraphAID` can be non-optional now, right?

LDSAnnotations/NSFileManager+URLs.swift
```diff
@@ -22,9 +22,9 @@
 
 import Foundation
 
-extension NSFileManager {
+public extension NSFileManager {
```
hiltonc: Let's remove this file from the framework if you agree with my assessment in `annotationStoreForUsername`.

LDSAnnotations/Session.swift
```diff
@@ -68,7 +68,7 @@ public class Session: NSObject {
     
     var lastSuccessfulAuthenticationDate: NSDate?
     
-    var authenticated: Bool {
+    public var authenticated: Bool {
```
hiltonc: Does this need to be public?

LDSAnnotations/SessionController.swift
```diff
@@ -0,0 +1,39 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+import Foundation
+
+public class SessionController {
+    
+    public static let sharedController = SessionController()
```
hiltonc: Same here, let's not make it a singleton in the framework.

LDSAnnotations/Bookmark.swift
stephanheilner: Still no, because we can have chapter level bookmarks (like when we add a doc to a notebook page). In that case we wouldn't have a paragraphAID. But this check is validating what is supposed to be a verse level paragraph (because it has a URI), that doesn't have an HTML5 version where the docmaps added a @pid. Without this check, it would think it was a valid chapter level bookmark, when in reality, it was tied to a specific paragraph.

LDSAnnotations/SessionController.swift
```diff
@@ -0,0 +1,39 @@
...
+    
+    private let cache = NSCache()
```
hiltonc: If you use NSCache, then sessions can disappear. Should it just be a dictionary?

LDSAnnotations/SyncAnnotationsOperation.swift
```diff
@@ -270,7 +270,8 @@ class SyncAnnotationsOperation: Operation {
                                 try annotationStore.addOrUpdateBookmark(downloadedBookmark)
                                 downloadBookmarkCount += 1
                             } else {
-                                throw Error.errorWithCode(.Unknown, failureReason: "Failed to deserialize bookmark: \(bookmark)")
+                                // Don't throw an error, this annotation will eventually come down when an HTML5 version becomes available
```
hiltonc: Can you make this (and the two following cases) throw in all cases except for specifically when the pid is missing but the URI is present? Maybe make the init for these objects throw instead of return nil, and throw 2 different errors for the two cases.

LDSAnnotations/Bookmark.swift
hiltonc: Got it. Thanks!

hiltonc:
This is great stuff.

LDSAnnotations/Session.swift
stephanheilner: I just changed my mind, but originally, I was wanting to use this to see if we have a valid username/password for the user. But really, this only tells us if they've signed in within the last few minutes, not if we have valid credentials, so I'll remove the public and do this a better way in the app.

LDSAnnotations/Error.swift
```diff
@@ -31,6 +31,12 @@ public struct Error {
         case AuthenticationFailed = -1001
         case LockedOut = -1002
         case PasswordExpired = -1003
+        case InvalidParagraphAID = -2000
```
stephanheilner: I wasn't sure if I should continue on with the -1004... or if I should start a new millenium of error codes :)

stephanheilner: I wasn't sure if I should continue on with the -1004... or if I should start a new millenium of error codes :)

hiltonc: hahahah, way more modern :)

LDSAnnotations/NSFileManager+URLs.swift
```diff
@@ -0,0 +1,31 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+import Foundation
+
+public extension NSFileManager {
```
hiltonc: Move this file back to the LDSAnnotationsDemo directory and make sure it's just in the demo and not the framework.

hiltonc:
👍

hiltonc:
That's cool that errors bubble up like that.

LDSAnnotations/AccountController.swift
nickmshelley: I agree that libraries shouldn't provide singletons. But I think they should definitely provide default/shared instances when it makes sense, like NSUserDefaults. I think it definitely makes sense here.

In fact, looking more closely at AccountController, it seems like the only reason we need an instance to begin with is for the observers; otherwise everything could easily be a class/static method. And since the observer sets are constants and not passed in on the init, we can safely remove them since they are guaranteed to be empty.

Or, if we want to fix the observers, we can make them static vars. I tried this in a playground and it worked:

``` swift
class MyClass {
    static var strings = [String]()
    
    static func printStrings() {
        strings.forEach { print($0) }
    }
}

MyClass.strings.append("hi")
MyClass.printStrings()
```

LDSAnnotations/SessionController.swift
nickmshelley: Same here, the properties and functions in this class should just be `static` or `class`.

PR #8, opened 2016-06-07
Add sharedController property to AccountController & SessionController

stephanheilner:


LDSAnnotations/SessionController.swift
```diff
@@ -24,16 +24,14 @@ import Foundation
 
 public class SessionController {
```
stephanheilner: I wonder if this would be better named `SessionCache` now, or something like that?

nickmshelley:
What's happening with this? Are you planning on changing anything or is this ready for review? I can't remember what we decided on this.

stephanheilner:
I have a few other changes I need to make. I think we stick with sharedControllers for the time-being. The changes in the current version of the PR aren't actually making any singletons, just using static vars and class functions, making it act like a singleton would. Stay tuned for another commit...

nickmshelley:
Sounds good.

stephanheilner:
If @hiltonc feels strongly about not including the sharedController static lets to the pod, then I can add the extra extensions to the demo project and GL to have the sharedController

hiltonc:
:+1: I think this is great. Thank you for not using statics other than the shared instances. I agree that nearly, if not all, library users will use that.

stephanheilner:
Thanks @hiltonc!

PR #9, opened 2016-06-09
Added functions for study notebook

stephanheilner:


LDSAnnotations/AnnotationStore+Tag.swift
```diff
@@ -79,6 +79,49 @@ extension AnnotationStore {
...
+    public func tagsOrderedByCount(withTagIDs tagIDs: [Int64]? = nil) -> [Tag] {
+        do {
+            let inClause: String
+            if let tagIDs = tagIDs {
+                inClause = String(format: "AND annotation_tag.tag_id IN (%@)", tagIDs.map({ String($0) }).joinWithSeparator(","))
+            } else {
+                inClause = ""
+            }
+            
+            let statement = "SELECT tag.* FROM tag, (SELECT annotation_tag.tag_id, COUNT(annotation_id) AS annotation_count FROM annotation_tag JOIN annotation ON annotation_tag.annotation_id = annotation._id WHERE annotation.status = '' \(inClause) GROUP BY tag_id) AS counts ON tag._id = counts.tag_id ORDER BY counts.annotation_count DESC, tag.name ASC"
```
stephanheilner: I tried for like 2 hours to make this work with SQLite.swift and then finally concluded that this was much simpler to do, and maintain. Feel free to correct me if I'm wrong.

stephanheilner:
... Still making a few changes, stay tuned

LDSAnnotations/AnnotationStore+Tag.swift
```diff
@@ -79,6 +79,44 @@ extension AnnotationStore {
...
+    }
+    
+    public func tagsOrderedByCount(withIDs ids: [Int64]? = nil) -> [Tag] {
+        do {
+            let inClause: String = {
+                guard let ids = ids else { return "" }
+                return String(format: "AND annotation_tag.tag_id IN (%@)", ids.map({ String($0) }).joinWithSeparator(","))
+            }()
+            
+            let statement = "SELECT tag.* FROM tag, (SELECT annotation_tag.tag_id, COUNT(annotation_id) AS annotation_count FROM annotation_tag JOIN annotation ON annotation_tag.annotation_id = annotation._id WHERE annotation.status = '' \(inClause) GROUP BY tag_id) AS counts ON tag._id = counts.tag_id ORDER BY counts.annotation_count DESC, tag.name ASC"
```
hiltonc: Yeah, once the query gets sufficiently crazy I fallback to SQL too. Maybe SQLite.swift can handle it, but diminishing returns.

stephanheilner:
Ok, its ready.

stephanheilner:
Ok, its ready now.

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -326,6 +326,22 @@ extension AnnotationStore {
         }
     }
     
+    public func annotationsWithNotebookIDCount(notebookID: Int64) -> Int {
```
hiltonc: Rename `numberOfAnnotationsWithNotebookID`

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -326,6 +326,22 @@ extension AnnotationStore {
...
+    
+    public func annotationsWithTagID(tagID: Int64) -> [Annotation] {
+        do {
+            return try db.prepare(AnnotationTable.table.join(AnnotationTagTable.table.filter(AnnotationTagTable.tagID == tagID), on: AnnotationTable.id == AnnotationTagTable.annotationID)).map { AnnotationTable.fromRow($0) }
+        } catch {
+            return []
+        }
+    }
+    
+    public func annotationsWithTagIDCount(tagID: Int64) -> Int {
```
hiltonc: `numberOfAnnotationsWithTagID`

LDSAnnotations/AnnotationStore+Notebook.swift
```diff
@@ -210,8 +201,42 @@ extension AnnotationStore {
...
+                query = query.filter(ids.contains(NotebookTable.id))
+            }
+            return try db.prepare(query).map { NotebookTable.fromRow($0) }
+        } catch {
+            return []
+        }
+    }
+    
+    /// Returns a list of active notebooks order by number of annotations in notebook descending.
+    public func notebooksOrderedByCount(ids ids: [Int64]? = nil) -> [Notebook] {
```
hiltonc: How about `notebooksOrderedByNumberOfAnnotations`?

LDSAnnotations/AnnotationStore+Notebook.swift
```diff
@@ -210,8 +201,42 @@ extension AnnotationStore {
...
+        let inClause: String = {
+            guard let ids = ids else { return "" }
+            
+            return String(format: "AND notebook._id IN (%@)", ids.map({ String($0) }).joinWithSeparator(","))
```
hiltonc: `ids.map { String($0) }.joinWithSeparator(",")`

LDSAnnotations/SyncAnnotationsOperation.swift
```diff
@@ -319,7 +319,7 @@ class SyncAnnotationsOperation: Operation {
                         
                         if let tags = rawAnnotation["tags"] as? [String: [String]] {
                             for tagName in tags["tag"] ?? [] {
-                                let downloadedTag = try Tag(name: tagName)
+                                let downloadedTag = try Tag(name: tagName.trimmed())
```
hiltonc: This makes me nervous. It seems like we should respect the tag as-is, and if we want this functionality push it to the backend. Otherwise all platforms have to implement it and if they don't there will be inconsistencies.

LDSAnnotations/Tag.swift
```diff
@@ -23,7 +23,7 @@
 import Foundation
 
 /// A Tag.
-public struct Tag {
+public class Tag {
```
hiltonc: Why class?

LDSAnnotations/Notebook.swift
```diff
@@ -23,7 +23,7 @@
 import Foundation
 
 /// A notebook.
-public struct Notebook {
+public class Notebook {
```
hiltonc: Why class?

LDSAnnotations/AnnotationStore+Tag.swift
```diff
@@ -79,6 +79,45 @@ extension AnnotationStore {
         }
     }
     
+    public func tags() -> [Tag] {
```
hiltonc: Can you combine these three into one func that takes an ordering enum (with a default)?

LDSAnnotations/AnnotationStore+Tag.swift
```diff
@@ -79,6 +79,45 @@ extension AnnotationStore {
...
+        do {
+            return try db.prepare(TagTable.table.select(distinct: TagTable.id, TagTable.name).join(AnnotationTagTable.table.join(AnnotationTable.table.select(AnnotationTable.id).filter(AnnotationTable.status == .Active), on: AnnotationTable.id == AnnotationTagTable.annotationID), on: TagTable.id == AnnotationTagTable.tagID)).map { TagTable.fromRow($0) }
+        } catch {
+            return []
+        }
+    }
+    
+    public func tagsOrderedByMostRecent(withIDs ids: [Int64]? = nil) -> [Tag] {
```
hiltonc: you can use do `ids ids` here and below to get closer to how Swift 3 naming will be.

LDSAnnotations/AnnotationStore+Notebook.swift
```diff
@@ -100,7 +91,7 @@ extension AnnotationStore {
             throw Error.errorWithCode(.Unknown, failureReason: "Cannot update a notebook without a name.")
         }
         
-        var modifiedNotebook = notebook
+        let modifiedNotebook = notebook
```
hiltonc: This pattern doesn't make sense any more if you're turning this into a class (here and elsewhere). But I really think you shouldn't make it a class.

hiltonc:
Also, can you write a unit test for each of these new methods?

LDSAnnotations/Tag.swift
```diff
@@ -36,7 +36,7 @@ public struct Tag {
         self.name = name
     }
     
-    init(name tagName: String?) throws {
+    public init(name tagName: String?) throws {
```
hiltonc: Why public?

LDSAnnotations/Notebook.swift
```diff
@@ -37,7 +37,7 @@ public struct Notebook {
     /// When the notebook was last modified in local time.
     public internal(set) var lastModified: NSDate
     
-    var uniqueID: String
+    public var uniqueID: String
```
hiltonc: Why public?

hiltonc:
Part of Tag and Notebook was made public, is that really needed?

LDSAnnotations/Tag.swift
stephanheilner: So that we can create a new tag from the app. The alternative is to add a function to AnnotationStore to add a tag by name. But everywhere else in AnnotationStore, we do addOrUpdate and take the object. We can't create the object we want to add unless this is public unless we change the way we want to insert records, which is totally fine. I just figured you had a reason you approached it the way you did.

LDSAnnotations/Notebook.swift
stephanheilner: We're not using URI anymore, so we need access to an identifier for notebooks. We could use the auto-increment ID, but that's not safe because it will be different on different devices (because its auto-incremented). `uniqueID` is the only unique identifier on Notebook. This is so we can store notebooks in the BackForwardHistory

stephanheilner:
I put explanations inline where you originally asked why they were public.

LDSAnnotations/Notebook.swift
hiltonc: Cool. It sounds like you don't need to modify it publicly. If so can you make it `internal(set)`?

hiltonc: Also, will you add a helpful doc comment to it, now that it's public?

hiltonc:
Thanks! I made the blanket comment because I wasn't sure if you would get notifications with the PR closed.

LDSAnnotations/Notebook.swift
stephanheilner: Will do, sounds good.

PR #10, opened 2016-06-10
Fixed missing table column

brandenr:
Fixed using incorrect table
Bump version

### DIFF
--- a/LDSAnnotations.podspec
+++ b/LDSAnnotations.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = "LDSAnnotations"
-  s.version      = "1.1.1"
+  s.version      = "1.1.2"
   s.summary      = "Swift client library for LDS annotation sync."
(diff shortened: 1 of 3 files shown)


hiltonc:
Can you add a quick little test for this? I think you can base it on `testAddNote` in https://github.com/CrossWaterBridge/LDSAnnotations/blob/master/LDSAnnotationsTests/AnnotationStoreTests.swift

hiltonc:
Also, for Travis CI to run tests you'll need to push your PR branch to this repo instead of your fork. This is because the tests use encrypted data, which Travis CI disables when building from forks. We can skip that for this, but it's something we can all do in the future.

nickmshelley:
Doesn't that mean that only committers can make PRs where all the tests will run? Or does it not matter since everyone that contributes is a committer anyway?

brandenr:
https://github.com/CrossWaterBridge/LDSAnnotations/pull/11

hiltonc:
@nickmshelley Yes, and yes.

PR #12, opened 2016-06-14
Added functions to select and add tags and notebooks

stephanheilner:


LDSAnnotations/Tag.swift
```diff
@@ -36,7 +36,7 @@ public struct Tag {
         self.name = name
     }
     
-    public init(name tagName: String?) throws {
+    init(name tagName: String?) throws {
```
hiltonc: I thought you were going to keep the init public!

hiltonc:
👍

LDSAnnotations/Tag.swift
stephanheilner: Changed my mind. :) This is actually easier to work with.

PR #13, opened 2016-06-30
Added ability to trash annotations

stephanheilner:
* Updated to CocoaPods 1.0
* Added functions to load notes, bookmarks, tags, etc


LDSAnnotations/Annotation.swift
```diff
@@ -52,17 +52,11 @@ public struct Annotation {
...
-
-    /// ID of associated bookmark. This is a one-to-one relationship
-    public var bookmarkID: Int64?
-    
     public var source: String?
     
     public var device: String?
     
-    init(id: Int64?, uniqueID: String, iso639_3Code: String, docID: String, docVersion: Int, type: AnnotationType, status: AnnotationStatus, created: NSDate?, lastModified: NSDate, noteID: Int64?, bookmarkID: Int64?, source: String?, device: String?) {
+    public init(id: Int64?, uniqueID: String, iso639_3Code: String, docID: String, docVersion: Int, type: AnnotationType, status: AnnotationStatus, created: NSDate?, lastModified: NSDate, source: String?, device: String?) {
```
nickmshelley: Why does this need to be public?

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -154,6 +146,19 @@ extension AnnotationStore {
         return modifiedAnnotation
     }
     
+    func updateLastModifiedDate(annotationID annotationID: Int64, status: AnnotationStatus? = nil) throws {
+        if let status = status {
+            try db.run(AnnotationTable.table.filter(AnnotationTable.id == annotationID).update(
+                AnnotationTable.lastModified <- NSDate(),
```
nickmshelley:
``` swift
try db.run(AnnotationTable.table.filter(AnnotationTable.id == annotationID).update(
  AnnotationTable.lastModified <- NSDate(),
  if let status = status {
    AnnotationTable.status <- status
  }
))

Comment (nickmshelley) [inline LDSAnnotations/AnnotationStore+Annotation.swift]: Oh wait, that probably doesn't work... never mind.

File: LDSAnnotations/Highlight.swift [hunk 3]
```diff
@@ -25,34 +25,38 @@ import Foundation
...
+    public static let OffsetEnd = -1
+    
     /// Local ID.
     public internal(set) var id: Int64?
     
     /// Paragraph Annotation ID.
     public let paragraphAID: String
     
-    /// The zero-based word offset of the start of this highlight range.
+    /// The word offset of the start of this highlight range.
```
nickmshelley: Instead of deleting "zero-based" here, can you just update it to be "crazy-based"?

LDSAnnotations/Highlight.swift
```diff
@@ -89,19 +84,10 @@ public struct Highlight: Equatable {
...
-        if let offsetEnd = offsetEnd {
-            result["@offset-end"] = offsetEnd + 1
-        } else {
-            result["@offset-end"] = -1
-        }
+        result["@offset-start"] = offsetStart > 1 ? offsetStart : Highlight.OffsetStart
+        result["@offset-end"] = offsetEnd ?? Highlight.OffsetEnd
         
-        if let style = style where style == .Underline {
+        if style == .Underline && style == .Clear {
```
nickmshelley: Isn't this guaranteed to be false?

LDSAnnotationsDemo.xcodeproj/project.pbxproj
```diff
@@ -1,453 +1,986 @@
-// !$*UTF8*$!
```
nickmshelley: This seems like a huge change for this file...

nickmshelley:
.

nickmshelley:
Why aren't tests running on Travis?

hiltonc:
Because this was pushed to a fork. The branch must be pushed to this report to trigger Travis because the tests use encrypted data.

nickmshelley:
Oh yeah, I feel like I knew that at one point. Can't remember if you told me or if it's documented somewhere.

LDSAnnotations/Annotation.swift
stephanheilner: I created an extension in the GL to initialize an Annotation with a subitem, but I can't create an instance of an Annotation unless the init is public

LDSAnnotations/Highlight.swift
stephanheilner: Oops, I meant `||`

LDSAnnotations/Highlight.swift
```diff
@@ -67,17 +71,8 @@ public struct Highlight: Equatable {
...
-            self.offsetStart = offsetStart - 1
-        } else {
-            self.offsetStart = 0
-        }
-        
-        if offsetEnd >= 1 {
-            self.offsetEnd = offsetEnd - 1
-        }
-        
+        self.offsetStart = offsetStart > 1 ? offsetStart : Highlight.OffsetStart
```
hiltonc: @stephanheilner Why did you go back to having clients use zero-based indexes? I thought it would be better to make things nicer for the client by encapsulating the eccentricities of the backend.

LDSAnnotations/Bookmark.swift
```diff
@@ -40,10 +42,10 @@ public struct Bookmark: Equatable {
     /// ID of annotation.
     public var annotationID: Int64
     
-    /// The zero-based word offset of this bookmark location.
-    public internal(set) var offset: Int
+    /// The word offset of this bookmark location.
+    var offset: Int
```
hiltonc: @stephanheilner Same here, I made it zero-based because the client deserves to have nice things. The library can encapsulate this.

LDSAnnotations/Highlight.swift
```diff
@@ -25,34 +25,38 @@ import Foundation
 /// A highlight.
 public struct Highlight: Equatable {
     
+    public static let OffsetStart = -1
```
hiltonc: As a user of this library, it's not clear to me what the difference between `OffsetStart` and `offsetStart` is. Same with `OffsetEnd` and `offsetEnd`.

LDSAnnotations/Highlight.swift
stephanheilner: Everything the app does is 1-based. So, in order to use this. So, every query, every lookup, every insert will have to convert it to zero-base. I didn't feel like it added enough value. I figured that the `fromRow` functions could easily be changed to convert to-from zero based, but if we do any lookups where we pass in the offset as a parameter, we'd have to remember to convert that. Or are you thinking the app would have to do the converting back and forth and the framework would just always be zero-based (except for when syncing to/from the server)?

LDSAnnotations/Highlight.swift
stephanheilner: Are you saying that there should be a command explaining, or that they should be named differently, or both?

hiltonc: Well, I kind of know what the difference is, but I'm not sure. I think I would recommend renaming the constants to be more clear.

LDSAnnotations/Highlight.swift
hiltonc: I think that everything the app does should be 0-based. So the conversion would only need to happen in this framework, when syncing to/from the server. Eventually even that can be pushed into the server, so that no client needs to do the conversion. I'm assuming that this is a longterm desirable goal.

I think we should apply this approach to all aspects of the data we get from or send to the server that seem a little weird. Any time we do one of these conversions, that's a great candidate for an improvement to the server later on.

PR #19, opened 2016-07-19
Added duplicateAnnotation function, fixed a bug

stephanheilner:


hiltonc:
:+1:, although I wonder if duplicateAnnotation shouldn't be in GL itself. I think you could implement it as a client, and I'm guessing it wouldn't generally be useful. But if you would prefer it to be here merge away.

stephanheilner:
I started that route, but didn't want to have to expose creating annotations to the app. I tried to make it so clients always create valid annotations with a note, tag, highlight, bookmark or link, but really you should never create an annotation without the related objects.

PR #20, opened 2016-07-20
Fix query to select tags to retain order correctly

stephanheilner:
This just makes the order tags by .MostRecent more efficient. The same tests still validate that it works correctly, just faster :)


hiltonc:
I love it when we can make refactorings like this with the confidence that tests offer.

PR #22, opened 2016-08-04
Fixed many sync bugs, and added sync tests

stephanheilner:
I also made a change to the data model to just derive the annotation type, instead of trying to persist it. It was causing so many sync problems because the type has to be changed depending on if the annotation has links, or highlights or is in a notebook. If we remove a link, for example, the type needs to be changed back to highlight. So, it was much simpler to just derive it when serializing the annotation.


LDSAnnotations/Annotation.swift
```diff
@@ -122,14 +111,19 @@ public struct Annotation: Equatable {
             result["note"] = note.jsonObject()
         }
 
+        // Derive the annotation type
+        var type: AnnotationType?
+        
         if let id = id, bookmark = annotationStore.bookmarkWithAnnotationID(id) {
```
nickmshelley: Not important, but you're doing `if let id = id` a few lines below, so the bookmark stuff could just go inside that one.

LDSAnnotations/Annotation.swift
```diff
@@ -139,23 +133,29 @@ public struct Annotation: Equatable {
...
             if !links.isEmpty {
-                result["references"] = ["reference": links.map({ $0.jsonObject() })]
+                result["refs"] = ["ref": links.map({ $0.jsonObject() })]
+                type = .Link // If there are 1 or more links, it MUST be a .Link type
             }
             
             let notebooks = annotationStore.notebooksWithAnnotationID(id)
             if !notebooks.isEmpty {
                 result["folders"] = ["folder": notebooks.map({ $0.annotationNotebookJsonObject() })]
+                type = .Journal
```
nickmshelley: Seems like this should go above the link stuff.

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -447,19 +440,22 @@ extension AnnotationStore {
...
     }
     
-    func deleteAnnotationWithID(id: Int64) {
-        do {
-            try db.run(AnnotationTable.table.filter(AnnotationTable.id == id).delete())
-            
-            notifySyncModifiedAnnotationsWithIDs([id])
-        } catch {}
+    func deleteAnnotationWithID(id: Int64) throws {
+        // Don't put this in a transaction, becuause it currently only called from within an existing transaction
```
nickmshelley: @hiltonc: This seems weird, but maybe just because I never understood the purpose of your `inTransaction` method...

LDSAnnotations/AnnotationStore+Highlight.swift
```diff
@@ -121,10 +121,22 @@ extension AnnotationStore {
         }
     }
     
-    func deleteHighlightWithID(id: Int64) {
-        do {
-            try db.run(HighlightTable.table.filter(HighlightTable.id == id).delete())
-        } catch {}
+    private func highlightWithID(id: Int64) -> Highlight? {
+        return db.pluck(HighlightTable.table.filter(HighlightTable.id == id)).map { HighlightTable.fromRow($0) }
```
nickmshelley: This could crash unless we use what's on SQLite.swift's master branch.

LDSAnnotations/AnnotationStore+Annotation.swift
hiltonc: Transactions should be reentrant, so it shouldn't hurt to wrap this in a transaction. But if it's really only called from within a transaction there's no need.

nickmshelley:
.

PR #23, opened 2016-08-05
Added notification for session added to SessionController

stephanheilner:


hiltonc:
Looks good, but what is this notification for? Because it's only called when you add a session, don't you already know that a session has been added?

PR #25, opened 2016-08-05
Make id property of Annotation not optional

stephanheilner:


LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -489,9 +440,11 @@ extension AnnotationStore {
 extension AnnotationStore {
     
     func notifyModifiedAnnotationsWithIDs(ids: [Int64]) {
+        
         let inSyncTransactionKey = "sync-txn:\(unsafeAddressOf(self))"
         guard NSThread.currentThread().threadDictionary[inSyncTransactionKey] == nil else {
-            fatalError("A local transaction cannot be started in a sync transaction")
+            // Don't send notifications while syncing
```
nickmshelley: @hiltonc: Can you take a look at this change?

nickmshelley: @stephanheilner: Do you know why the fatalError started being triggered? This PR doesn't seem like it should affect when notifications are sent, so it seems suspicious that removing the fatalError was necessary at all.

nickmshelley: So it looks like those notify methods batch up notifications if it's in a transaction, and there's a separate one for sync that seems identical except for which transaction key is used. If there's not an easy way to just put them all in the same bucket regardless of what kind of transaction we're in and then bulk notify at the end, then you can get around that problem by just passing an `inSync` parameter to `addAnnotation` that will tell it to call the right method. That will prevent notifications from being lost.

hiltonc: I think we should keep the fatalError. There should be a clear separation between transactions initiated by the user and transactions initiated by a sync. It's OK for them to be happening concurrently, but this fatalError catches the case where a user transaction starts running sync code, or sync code starts running a user change.

nickmshelley: Also, it seems weird that `trashAnnotationIfEmptyWithID` calls `notifySyncModifiedAnnotationsWithIDs`. Why is it always calling the sync one? Don't we only call that method outside of sync?

hiltonc: The purpose of the notifications is so that the app can be notified when there are new changes from a user action vs. a sync. The app might want to trigger a sync when a user action happens. The app would not want to trigger a sync in response to a sync happening.

nickmshelley: Ok that makes sense. I didn't see the `source` on the notify at first.

nickmshelley: I'd prefer we change it to an `assert` though. The user getting notified of the wrong type of action doesn't seem fatal (a.k.a. crash-worthy) to me.

nickmshelley:
.

LDSAnnotations/AnnotationStore+Annotation.swift
hiltonc: It represents a major flaw in the library. I wonder if we can use the type system to make such errors impossible? I'd prefer this remain a fatalError, but assert should still make the point.

stephanheilner: So, before, we had a func called `addOrUpdateAnnotation` and it used the `notifySyncModifiedAnnotationsWithIDs([id])`, but I removed that and started using  `addAnnotation` or `updateAnnotation`, both of which use `notifyModifiedAnnotationsWithIDs`. So, I guess we could pass in a Bool value to the addAnnotation and updateAnnotation to indicate if its syncing and call current notify func?

stephanheilner: Or maybe we could just have a single func we call, that is smart enough to know if its mid-sync and do the right thing?

hiltonc: Any public functions shouldn't have a sync flag. But any internal methods could, which would let us use them for either case.

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -98,13 +98,17 @@ extension AnnotationStore {
...
+        } else {
+            notifyModifiedAnnotationsWithIDs([id])
+        }
         
         return Annotation(id: id, uniqueID: uniqueID, iso639_3Code: iso639_3Code, docID: docID, docVersion: docVersion, status: .Active, created: created, lastModified: lastModified, source: source, device: device)
     }
     
     /// Saves any changes to `annotation` and updates the `lastModified`.
-    public func updateAnnotation(annotation: Annotation) throws -> Annotation {
+    public func updateAnnotation(annotation: Annotation, inSync: Bool = false) throws -> Annotation {
```
nickmshelley: This parameter shouldn't be on a public func.

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -77,7 +77,7 @@ extension AnnotationStore {
     }
     
     /// Adds a new annotation.
-    func addAnnotation(uniqueID uniqueID: String? = nil, iso639_3Code: String, docID: String?, docVersion: Int?, status: AnnotationStatus = .Active, created: NSDate? = nil, lastModified: NSDate? = nil, source: String, device: String) throws -> Annotation {
+    func addAnnotation(uniqueID uniqueID: String? = nil, iso639_3Code: String, docID: String?, docVersion: Int?, status: AnnotationStatus = .Active, created: NSDate? = nil, lastModified: NSDate? = nil, source: String, device: String, inSync: Bool = false) throws -> Annotation {
```
nickmshelley: Part of me doesn't want to give `inSync` a default parameter so you're forced to think about it every time you use the function (which may help avoid bugs), but it seems like there are only two places that pass `true` so I'm torn...

stephanheilner: I know, right? But, I thought it would be okay since its not public anyway.

nickmshelley: Yeah it's only `true` in 2 places so I could go either way.

nickmshelley:
👍  from me, but I think @hiltonc should weigh in too since he was part of the review.

hiltonc:
Beauty!

hiltonc:
👍

PR #27, opened 2016-08-08
Make all primary key id: Int64 non-optional values

stephanheilner:


hiltonc:
This looks good to me.

nickmshelley:
👍

PR #28, opened 2016-08-09
Skip annotations that have invalid highlights, links, notes, etc

stephanheilner:
Before we we skipping just pieces of the annotation. For example, if the highlight didn't have a @pid, we just skipped the highlight, but did save the annotation and any notes in the db. That's not what we want though, because it lets us change the annotation. If an annotation is invalid (which should in pretty much all cases be from it not being updated for HTML5), we just skip that annotation. When HTML5 becomes available for that annotation, it will come down from the service as a change and be valid and saved.

### DIFF
--- a/LDSAnnotations/AnnotationStore.swift
+++ b/LDSAnnotations/AnnotationStore.swift
@@ -151,6 +151,5 @@ public class AnnotationStore {
             }
         }
     }
-
     
 }
--- a/LDSAnnotations/Error.swift
+++ b/LDSAnnotations/Error.swift
@@ -41,6 +41,7 @@ public struct Error {
         case SaveHighlightFailed = -3001
         case RequiredFieldMissing = -3008
         case SyncFailed = -4000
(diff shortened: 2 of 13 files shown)


SerenadeX:
Quite a lot going on here but it looks like most everything was wrapped up in a `do {} catch {}` to handle the new `deserializationErrors` that you put in.  Is that right?

stephanheilner:
I removed some do/catches, added a savepoint within the transaction, and separated out all the deserialization errors so that each one would be descriptive and then instead of failing on an error, keep an array of all the deserialization errors (and fail on any other kind of error) and pass the deserialization errors through in the success result so the client can do something with it (like log to analytics) if it wants to. There's not much new, just mostly refactoring. I wish it would ignore whitespace so the diff didn't look so scary.

SerenadeX:
The nice thing is is it makes it darker green if its new.  Looks like mostly refactoring.

SerenadeX:
👍

PR #29, opened 2016-08-10
Added notifications for auth, deserialize & sync failures

stephanheilner:


stephanheilner:
Going to just do this in GL

PR #31, opened 2016-08-11
Added public var for last local sync date

stephanheilner:


LDSAnnotations/SyncToken.swift
```diff
@@ -87,4 +88,9 @@ public struct SyncToken {
         return "1:\(localSyncNotebooksDate.timeIntervalSince1970):\(serverSyncNotebooksDate.timeIntervalSince1970):\(localSyncAnnotationsDate.timeIntervalSince1970):\(serverSyncAnnotationsDate.timeIntervalSince1970)"
     }
     
+    /// Last sync date
+    public var lastLocalSyncDate: NSDate {
```
hiltonc: I was hoping we could keep this opaque, but I'm guessing you want it for display to the user?

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -337,7 +337,7 @@ extension AnnotationStore {
     }
     
     /// Returns a the number of active annotations in notebook with ID
-    public func numberOfAnnotations(notebookID notebookID: Int64) -> Int {
+    public func numberOfAnnotationsInNotebook(notebookID: Int64) -> Int {
```
hiltonc: The old name seemed a little more Swift 3ish. This is fine too, just observing.

hiltonc:
👍 just a few observations.

LDSAnnotations/SyncToken.swift
stephanheilner: I actually need it for 2 things in the Gospel Library. 
1) The prompt to get user to sign in if they have unsynced annotations
2) The support email displays how many unsynched annotations they have. 
In order to get the uncynched count, we need to last synced date. We COULD pass it in the completion block (we need the exact timestamp used in sync, not just when sync finished), or we could use the sync token. Whaddayathink?

LDSAnnotations/AnnotationStore+Annotation.swift
hiltonc: In my previous comment I was saying that this function used to be able to do more than it does now. Before you could pass in nil (or just not supply the notebookID) and it would give you a count of all annotations, not just those in a specific notebook. I was asking if that was not useful functionality?

LDSAnnotations/SyncToken.swift
hiltonc: I think what you have is A-OK.

stephanheilner:
Tests will fail with this latest commit, because the services team rolled up an update that doesn't allow us to send up bogus docIDs. I will fix that in a separate PR.

PR #32, opened 2016-08-11
Removed iso639_3 code, service is taking care of that now

stephanheilner:
* Also use valid docIDs in tests so service doesn't reject annotations


hiltonc:
Yay deleted code!

PR #33, opened 2016-08-16
Fix two join queries

stephanheilner:


hiltonc:
This looks good. Can you add a simple unit test for each of these methods?

PR #34, opened 2016-08-25
Use transactions for database inserts, update and deletes

stephanheilner:
* Moved all public AnnotationStore functions in their own extensions


nickmshelley:
Travis is having trouble with Mac builds, so I'll trust Stephan that tests pass and merge.

PR #35, opened 2016-09-01
Configure db to retry if its locked

stephanheilner:


LDSAnnotations/AnnotationStore.swift
```diff
@@ -51,7 +51,11 @@ public class AnnotationStore {
     public init?(path: String? = nil) {
         do {
             db = try Connection(path ?? "")
-            
+            db.busyHandler { _ in
```
nickmshelley: Just curious why we're doing a handler instead of a timeout like we do elsewhere. Is it just to make sure all queries get executed no matter how long they take?

nickmshelley:
👍 One question, but feel free to merge.

LDSAnnotations/AnnotationStore.swift
stephanheilner: Yeah, I thought about the timeout, but I'm not sure how many times it retries, and when it fails. For us, really we just want to keep retrying the query until the db is unlocked.

PR #36, opened 2016-09-08
Use common var for notification keys

stephanheilner:


LDSAnnotationsTests/XCTestCase+LDSAnnotations.swift
```diff
@@ -63,7 +63,7 @@ extension XCTestCase {
                     XCTFail("Failed with errors \(errors)")
                 }
             }
-            NSThread.sleepForTimeInterval(1)
+            NSThread.sleepForTimeInterval(2) // Add 2 second delay between syncs to be sure their trigger has finished
```
hiltonc: Can you explain what you mean by "their trigger has finished"? Also, better to not mention 2 in the comment because later someone will change the sleep but not update the comment.

LDSAnnotations/AnnotationStore.swift
```diff
@@ -45,6 +45,14 @@ public class AnnotationStore {
         return "local-txn:\(unsafeAddressOf(self))"
     }()
     
+    lazy var annotationNotificationKey: String = {
```
hiltonc: Can you just encapsulate the whole `NSThread.currentThread().threadDictionary` mechanism instead?

hiltonc:
Beauty!!

hiltonc:
👍

PR #39, opened 2016-09-09
Only allow one sync operation at a time.

nickmshelley:
Add a grace period between syncs.

Limit queued syncs to one per session.


LDSAnnotations/Session.swift
```diff
@@ -86,6 +86,8 @@ public class Session: NSObject {
     
     var lastSuccessfulAuthenticationDate: NSDate?
     
+    private var syncQueued = false
```
hiltonc: I think you need to protect access to this state because sync can be called on different threads.

LDSAnnotations/SyncAnnotationsOperation.swift
```diff
@@ -23,11 +23,16 @@
...
 import Operations
 
-class SyncAnnotationsOperation: Operation {
+enum SyncAnnotationsOperationError: ErrorType {
+    case NotebookSyncFailed(String)
+}
+
+class SyncAnnotationsOperation: Operation, AutomaticInjectionOperationType {
+    
+    var requirement: SyncNotebooksResult?
```
hiltonc: This seems like an odd name, is it a term of art?

LDSAnnotations/SyncOperation.swift
```diff
@@ -0,0 +1,53 @@
...
+                    downloadBookmarkCount: syncAnnotationsChanges.downloadBookmarkCount,
+                    downloadHighlightCount: syncAnnotationsChanges.downloadHighlightCount,
+                    downloadTagCount: syncAnnotationsChanges.downloadTagCount,
+                    downloadLinkCount: syncAnnotationsChanges.downloadLinkCount)
+                completion(SyncResult.Success(token: token, changes: changes, deserializationErrors: (notebooksResult.deserializationErrors + syncAnnotationsDeserializationErrors)))
+            case let .Error(errors: errors):
+                completion(SyncResult.Error(errors: errors))
+            }
+            
+            // The annotation service team asked that we have a grace period between syncs. Just sleeping at the end of this mutually exclusive operation provides that functionality without holding anything else up (since the completions have already been called)
```
hiltonc: Or they could fix their service instead of relying on clients to be well behaved...

hiltonc: Just grumping, don't mind me.

LDSAnnotations/Session.swift
nickmshelley: Well the risk is that we do an extra sync that syncs nothing, so I figured that simplicity was more important here. But I'm fine with guaranteeing correctness if you want. Lemme know.

LDSAnnotations/SyncAnnotationsOperation.swift
nickmshelley: Yes, if you consider protocol conformance art. :)

LDSAnnotations/SyncOperation.swift
nickmshelley: Agreed.

LDSAnnotations/Session.swift
nickmshelley: Actually, I'm legitimately curious if it's possible to have threading issues when setting a bool. I think that adding sets to a serial queue would be identical to just having each thread set it itself, since both seem like first-come-first-serve behavior.

LDSAnnotations/SyncOperation.swift
stephanheilner: Totally dumb, I know. But tests kept failing until we added the delay. But in the real world, people aren't really syncing every second so I guess they though it was acceptable. I'm still working on them...

LDSAnnotations/SyncAnnotationsOperation.swift
hiltonc: Oh. I will have to become acquainted with this library :)

LDSAnnotations/Session.swift
hiltonc: It's a little difficult for me to analyze. I think you're right that there is no actual problem with this code. If you had the whole method in a dispatch_sync, and the whole completion of the SyncOperation in a dispatch_sync, then it would be much easier to reason about. If you would prefer not to, then I'm OK with it.

hiltonc:
👍

LDSAnnotations/Session.swift
nickmshelley: Lemme look at it more before it's merged.

nickmshelley: I'm going to synchronize access, but I can't seem to find anything that says whether or not `addOperation` is thread safe. Now I'm worried we need to do the same thing in LDSContent.

And now after actually looking at it, it looks like we have some progress and completion dictionaries that aren't synchronized...

LDSAnnotations/SyncNotebooksOperation.swift
```diff
@@ -119,10 +102,10 @@ class SyncNotebooksOperation: Operation {
             return result
         }
         
-        return localChanges
+        return (localChanges, modifiedNotebooks)
     }
     
-    func applyServerChanges(payload: [String: AnyObject], onOrBefore: NSDate) throws {
+    func applyServerChanges(payload: [String: AnyObject], onOrBefore: NSDate) throws -> (NSDate, [String: [String]], [Notebook], [NSError]) {
```
stephanheilner: Can you make `[NSError]`, `[ErrorType]`?

LDSAnnotations/SyncNotebooksOperation.swift
```diff
@@ -131,12 +114,11 @@ class SyncNotebooksOperation: Operation {
             throw Error.errorWithCode(.Unknown, failureReason: "Missing before")
         }
         
-        self.serverSyncNotebooksDate = serverSyncDate
-        
         var notebookAnnotationIDs = [String: [String]]()
+        var downloadedNotebooks = [Notebook]()
+        var deserializationErrors = [NSError]()
```
stephanheilner: Can you make `[NSError]`, `[ErrorType]`?

stephanheilner: It doesn't bridge them for free without a cast, so this could actually cause a crash.

nickmshelley: No crashes that the tests caught at least. :)

stephanheilner: True... maybe it does work, but I was looking at some crash logs and it looked like that was the root cause. Dunno... but simple enough to change and not worry about he risk, right?

nickmshelley: Yeah I changed it before making the comment. I was thinking more that we should add tests if we can reproduce crashes.

nickmshelley:
@stephanheilner Was there something else or is this good to merge?

PR #41, opened 2016-09-13
Fixed several sync bugs based on load test.

stephanheilner:


stephanheilner:
The 64-bit Int.max was causing sync to fail because it can't handle such a large number.

LDSAnnotations/SyncAnnotationsOperation.swift
```diff
@@ -50,6 +50,7 @@ class SyncAnnotationsOperation: Operation {
     var deserializationErrors = [NSError]()
     
     private let source: NotificationSource = .Sync
+    private let awsMaxInt = 99999
```
hiltonc: To me AWS == Amazon Web Services. Can you call it sync service or something like that?

LDSAnnotations/AnnotationStore+Bookmark.swift
```diff
@@ -132,11 +132,11 @@ extension AnnotationStore {
             // Mark all the annotations that were updated as changed
             try annotationIDs.forEach { try self.updateLastModifiedDate(annotationID: $0, source: source) }
             
-            return try self.addBookmark(name: name, paragraphAID: paragraphAID, displayOrder: displayOrder, annotationID: annotation.id, source: source)
+            return try self.addBookmark(name: name, paragraphAID: paragraphAID, displayOrder: displayOrder, annotationID: annotation.id, offset: Bookmark.DisplayOrderMax, source: source)
```
hiltonc: This seems like a mismatch. Why are we passing `DisplayOrderMax` to `offset`?

LDSAnnotations/Bookmark.swift
```diff
@@ -65,7 +66,7 @@ public struct Bookmark: Equatable {
         if let paragraphAID = paragraphAID {
             result["@pid"] = paragraphAID
         }
-        if let displayOrder = displayOrder {
+        if let displayOrder = displayOrder where displayOrder != Bookmark.DisplayOrderMax {
```
hiltonc: Why does `displayOrder` take two special values (`nil` and `DisplayOrderMax`), can't it just use `nil` to represent "no display order"?

stephanheilner: `.order(` supports using `??`, so I'm going to follow your suggestion and use `?? Int.max`, which should put the nulls at the end.

LDSAnnotations/SyncNotebooksOperation.swift
```diff
@@ -178,6 +177,16 @@ class SyncNotebooksOperation: Operation {
                                 downloadedNotebook = try self.annotationStore.addNotebook(uniqueID: uniqueID, name: name, description: description, status: status, lastModified: lastModified, source: self.source)
                             }
                             downloadedNotebooks.append(downloadedNotebook)
+                            
+                            if let annotationUniqueIDs = annotationUniqueIDs where !annotationUniqueIDs.isEmpty {
```
hiltonc: This can be simplified to:

    `for annotationUniqueID in annotationUniqueIDs ?? [] {`

stephanheilner:
I'm not 100% sure it works to be honest, but I was going through the source to see how to create an `IFNULL` Expression, and then found they have a function for `??` for the IFNULL function in SQLite.

stephanheilner:
Okay to merge?

hiltonc:
There's the one last recommendation if you want to do it. 👍 otherwise.

stephanheilner:
Oh right, thought I'd already pushed that change.

PR #42, opened 2016-10-07
Fix multiple re-syncing

stephanheilner:


nickmshelley:
👍  Thanks for fixing my crap! 😄

LDSAnnotationsTests/SessionNotificationsTests.swift
```diff
@@ -30,29 +30,6 @@ class SessionNotificationsTests: XCTestCase {
         XCTAssertEqual(session.status, Session.Status.Unauthenticated)
     }
     
-    func testAuthSuccessNotifications() {
```
hiltonc: We dropped a test here, would it make sense to replace it, or is it no longer needed?

stephanheilner:
The two that I dropped we to test the status notifications when sign in fails and succeeds. Since we don't update the status for those anymore, the tests wouldn't be helpful. If we add an "authStatus" with observers, then we'll want them again.

hiltonc:
OK, cool. Thanks!

PR #45, opened 2016-10-12
(GLI-1959) Fixed network activity indicator not showing when syncing

skytan4:


hiltonc:
This won't play well with all apps. For example, in many apps the app maintains a count of times a network event started and ended. When it is > 0, it shows the indicator.

Instead of this, can you provide a hook for an app to update it's own indicator?

PR #46, opened 2016-10-13
(GLI-1959) Add network indicator hooks

skytan4:


hiltonc:
The names networkIndicatorStart and networkIndicatorStop assume that this is only useful for starting and stopping the network indicator. In the case where an app is tracking the number of network requests outstanding and updating the network indicator accordingly, these names are inaccurate.

As a suggestion, but I'm not saying this is the best way to do it, consider creating a networkActivityObservers: ObserverSet<NetworkActivity> on session, and access it through the session property of the operation (instead of at the static level with Session).

nickmshelley:
I actually like it better as static methods because then you just set it once and forget it instead of having to make sure it's set for each session change. I think @stephanheilner suggested observers too, though, so maybe I'm outnumbered...

skytan4:
I've got both versions in two commits so we're good to go with whatever is chosen

nickmshelley:
Sorry I meant to comment earlier but something came up before I could press send.

hiltonc:
network activity = Active / Inactive

hiltonc:
or whatever. But not Increment/Decrement

nickmshelley:
I think you have `Increment` and `Decrement` backwards in the current version.

skytan4:
Sorry I thought active/inactive was similar to start/stop, but I see that now,  I can switch it @hiltonc. @nickmshelley I was having the counter increment when data task resumes/begins and then in the closure decrement/stop it like we talked about earlier today.

nickmshelley:
I talked with Hilton and I'm fine with observers, especially considering that's what @stephanheilner suggested to begin with. So proceed with that.

nickmshelley:
Leaving this one to @hiltonc now.

hiltonc:
Wait, @nickmshelley is leaving this PR to me? Alright! Skyler, let's make some massive changes that @nickmshelley won't like. I'll get back to you with my list.

LDSAnnotations/Session.swift
```diff
@@ -51,12 +57,22 @@ public class Session: NSObject {
     let authenticationURL: NSURL?
     let domain: String
     let trustPolicy: TrustPolicy
+    
+    static var networkIndicatorStart: (() -> Void)?
```
hiltonc: These two vars are unused.

LDSAnnotations/Session.swift
```diff
@@ -153,6 +168,7 @@ public class Session: NSObject {
 extension Session {
     
     func put(endpoint: String, payload: [String: AnyObject], completion: (Response) -> Void) {
+        networkActivity = .Inactive
```
hiltonc: It is possible to have two or more network tasks in flight at the same time. As a consequence, depending on timing the state can be incorrect.

One way to solve this is to track the number of outstanding network tasks at any given time and update the networkActivity accordingly.

nickmshelley: @hiltonc Are you saying go back to the static closure approach?

It seems like this current approach will work fine as long as we don't think of the enum as a state. So since the enum is called `NetworkActivity`, maybe just rename the cases to `Start` and `Stop`?

hiltonc: Technically that would work, but the idea of an observer is to observe state. I'd prefer to use a mechanism that matches the idea.

hiltonc: Or an idea that matches the mechanism.

nickmshelley: Oh I see. So you don't think it makes sense for observers to fire without a backing var? I was thinking just get rid of the var and fire the observers. It seems like we do that with sync updates (there's no var backing what was updated in sync, but we still send observer notifications), so I'm not really seeing how this is different...

But this is your PR, I'm just backseat driving. :)

hiltonc: Yeah, that's a good point. I think those are also being used incorrectly :). But given precedent, for now we could just rename these Start/Stop like you say.

skytan4: 😃 Thanks @hiltonc !, although I'll do what is asked of me if needs be.

LDSAnnotations/AuthenticateOperation.swift
```diff
@@ -115,6 +116,7 @@ class AuthenticateOperation: Operation {
                 self.finish(Error.errorWithCode(.Unknown, failureReason: "Authentication for an unknown reason."))
             }
         }
+        self.session.networkActivityObservers.notify(.Start)
```
hiltonc: self not needed

LDSAnnotations/Session.swift
```diff
@@ -153,6 +159,7 @@ public class Session: NSObject {
 extension Session {
     
     func put(endpoint: String, payload: [String: AnyObject], completion: (Response) -> Void) {
+        networkActivityObservers.notify(.Stop)
```
hiltonc: Need to fire .Stop when network activity stops, not before starting the network request

nickmshelley:
Tests passed for @skytan4 locally.

PR #47, opened 2016-10-24
Annotation with only clear highlights should be trashed

stephanheilner:


LDSAnnotationsTests/AnnotationStoreTests.swift
```diff
@@ -756,6 +756,30 @@ class AnnotationStoreTests: XCTestCase {
...
+        try! annotationStore.trashNoteWithID(note.id)
+        
+        // Verify note has been deleted
+        XCTAssertNil(annotationStore.noteWithID(note.id))
+        
+        // Verify no annotations are associated with note
+        XCTAssertNil(annotationStore.annotationWithNoteID(note.id))
+        
+        // Verify annotation is trashed
+        XCTAssertTrue(annotationStore.annotationWithID(annotation.id)?.status == .Trashed)
```
hiltonc: Just in case, can you add a similar test, but for a non-clear highlight, to make sure that it is not trashed?

nickmshelley:
Stephan showed me that his tests pass locally. I assume @hiltonc is fine with the existing test, but if not we can add one in a new PR.

LDSAnnotationsTests/AnnotationStoreTests.swift
hiltonc: No, that's great. I didn't realize that.

PR #50, opened 2016-10-27
Don't overwrite last modified dates on sync.

nickmshelley:
Notify of changes once at end of sync.


nickmshelley:
Why are github's diffs so horrible?

nickmshelley:
Here's what the diff looks like in Xcode.
screen shot 2016-10-27 at 4 51 48 pm

nickmshelley:
Tests passed locally.

stephanheilner:
👍 

The pre-load existing annotations will be in a follow-on commit?

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -333,7 +333,10 @@ extension AnnotationStore {
                 AnnotationTable.device <- device
             ))
             
-            try self.notifyModifiedAnnotationsWithIDs([id], source: source)
+            // Sync notifies once at the end
+            if case .Local = source {
```
stephanheilner: I've never seen this syntax before. Pretty slick!

nickmshelley:
Yeah I figured it would be best to get the modified date bug fix in sooner rather than later.

PR #53, opened 2016-11-01
Add ability to merge one annotation store into another.

nickmshelley:


LDSAnnotations/AnnotationStore+Merge.swift
```diff
@@ -0,0 +1,66 @@
...
+        let notebooks = otherStore.notebooks()
+        for notebook in notebooks {
+            do {
+                try addNotebook(uniqueID: notebook.uniqueID, name: notebook.name, description: notebook.description, status: notebook.status, lastModified: notebook.lastModified, source: .Sync)
+            } catch {
+                NSLog("Unable to add notebook: \(error)")
+            }
+        }
+        
+        for annotation in otherStore.annotations() {
```
stephanheilner: Another option would be to use the function called `duplicateAnnotation`, refactor it to take an annotationStore and then you don't have to duplicate all this logic?

nickmshelley:
Tests pass locally.

stephanheilner:
Suggestion or 👍

LDSAnnotations/AnnotationStore+Merge.swift
```diff
@@ -0,0 +1,66 @@
...
+//
+//  Created by Nick Shelley on 11/1/16.
+//  Copyright © 2016 Hilton Campbell. All rights reserved.
+//
+
+import Foundation
+
+extension AnnotationStore {
+    
+    public func mergeWith(otherStore: AnnotationStore, appSource: String, device: String) {
```
hiltonc: I think the name could be a little clearer. Maybe: addAnnotations(from otherStore: AnnotationStore, appSource: String, device: String)

nickmshelley: I like that better, except it also adds notebooks, and I don't really like `addAnnotationsAndNotebooks(from...)`. Maybe `addAll(from...)`? I dunno.

nickmshelley:
Tests pass locally.

PR #58, opened 2016-12-02
Hide chapter bookmarks

stephanheilner:


stephanheilner:
Tests passed locally. So I'll merge.

PR #59, opened 2016-12-05
Explicitly set obssocookie in request headers

stephanheilner:


LDSAnnotations/Session.swift
```diff
@@ -58,6 +58,14 @@ public class Session: NSObject {
     let domain: String
     let trustPolicy: TrustPolicy
     
+    static let ObSSOCookieName = "ObSSOCookie"
+    var obSSOCookie: String?
+    
+    public var obSSOCookieHeader: (name: String, value: String)? {
```
stephanheilner: This is public so we can use it for secure catalogs.

hiltonc: It would be good to move auth out into a shared library some time in the future.

LDSAnnotations/Session.swift
```diff
@@ -58,6 +58,14 @@ public class Session: NSObject {
     let domain: String
     let trustPolicy: TrustPolicy
     
+    static let ObSSOCookieName = "ObSSOCookie"
```
hiltonc: This is the name of the cookie, but can you give the variables you use a more descriptive name? Like `sessionCookie`?

stephanheilner:
Tests pass locally.

LDSAnnotations/Session.swift
stephanheilner: for sure.

PR #60, opened 2016-12-06
Don't make versioned annotation request if we have no ids

stephanheilner:
This also fixes a bug where it doesn't update the syncToken when it adds a sync to the queue because something changed mid-sync.


stephanheilner:
All tests passed locally.

PR #62, opened 2016-12-13
Don't overwrite lastModified date in updateAnnotation.

nickmshelley:


nickmshelley:
The only two places that all `updateAnnotation` are sync and `updateBookmark`. I verified that bookmark changes still sync across devices as they should. And I also verified that we keep the correct `lastModified` when syncing down changes to existing annotations.

stephanheilner:
:+1:

nickmshelley:
Tests passed locally, but I should probably add a new test. However, I want this in GL before we ship 4.0 (hopefully), so I'll make a task to write the test.

PR #64, opened 2017-01-05
Fix performance of check to see if annotation is empty

stephanheilner:
* I profiled adding/removing tags from an annotation, and it showed that this selecting the count from multiple tables was taking some time. So I optimized it into a single query. Instruments showed a significant enough difference after the change
* Added an index on the tag.name column
* Renamed enum values of HighlightStyle to match swift3 convention.


LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -452,14 +452,12 @@ extension AnnotationStore {
...
         try inTransaction(notificationSource: source) {
-            let notEmpty = [
-                try self.db.scalar(HighlightTable.table.filter(HighlightTable.annotationID == id && HighlightTable.style != .Clear).count),
-                try self.db.scalar(LinkTable.table.filter(LinkTable.annotationID == id).count),
-                try self.db.scalar(AnnotationTagTable.table.filter(AnnotationTagTable.annotationID == id).count),
-                try self.db.scalar(BookmarkTable.table.filter(BookmarkTable.annotationID == id).count),
-                try self.db.scalar(NoteTable.table.filter(NoteTable.annotationID == id).count)
-            ].any { $0 > 0 }
-            
+            let statement = "select count(annotation_id) from highlight where annotation_id = @annotationID and style != @style UNION ALL select count(annotation_id) from link where annotation_id = @annotationID UNION ALL select count(annotation_id) from annotation_tag where annotation_id = @annotationID UNION ALL select count(annotation_id) from bookmark where annotation_id = @annotationID UNION ALL select count(annotation_id) from note where annotation_id = @annotationID;"
```
stephanheilner: Sqlite.swift doesn't support union yet :(

LDSAnnotations.xcodeproj/project.pbxproj
```diff
@@ -680,7 +680,7 @@
 				ONLY_ACTIVE_ARCH = YES;
 				SDKROOT = iphoneos;
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
-				SWIFT_VERSION = 3.0.1;
+				SWIFT_VERSION = 3.0;
```
stephanheilner: These were to get LDSAnnotations compiling on Xcode 8.2

stephanheilner:
Tests passed locally.

hiltonc:
👍  nice profiling and fix

PR #68, opened 2017-01-11
Add associated link to link destination case.

nickmshelley:


nickmshelley:
Tests pass locally, but don't merge yet because I accidentally made the PR before making sure it works as expected in GL.

nickmshelley:
Ok this is working like I want it to. Please review. Kthxbye!

LDSAnnotations/AnnotationType.swift
```diff
@@ -21,11 +21,38 @@
...
+            return true
+        } else {
+            return false
+        }
+    }
+}
+
+public func ==(lhs: AnnotationType, rhs: AnnotationType) -> Bool {
+    switch (lhs, rhs) {
+    case (.link, .link):
```
stephanheilner: This is okay, as-is. But what do you think of changing it to something like this?
```
public func ==(lhs: AnnotationType, rhs: AnnotationType) -> Bool {
    switch (lhs, rhs) {
    case (.link, .link), (.multiple, .multiple), (.note, .note), (.notebook, .notebook), (.tag, .tag):
        return true
    case let (.linkDestination(leftLink), .linkDestination(rightLink)):
        return leftLink == rightLink
    default:
        return false
    }
}
```

PR #70, opened 2017-03-01
Have addAll reset lastModified so sync will pick them up.

nickmshelley:
This isn't the ideal solution, but it's the simplest, and given the rarity of the merge annotations situation we felt like simple was better than "correct."


nickmshelley:
Tests pass locally.

PR #71, opened 2017-03-16
Don't set invalid value of 'clear' for @style, created HighlightColor enum

stephanheilner:
… enum


stephanheilner:
Tests pass locally (when running against the prod env). Sign in for the beta lane is down right now.

LDSAnnotations/AccountController.swift
```diff
@@ -84,4 +87,8 @@ public class AccountController {
         UserDefaults.standard.set(tokens, forKey: AccountController.tokensKey)
     }
     
+    public func deleteSyncToken(forUsername username: String) {
+        UserDefaults.standard.removeObject(forKey: username)
```
nickmshelley: Doesn't look like this is doing the right thing.

stephanheilner: I did this incorrectly, pushing another commit with a fix.

hiltonc: It looked funny to me at first, so I looked at where we store the token, but didn't realize the two didn't match.

LDSAnnotations/HighlightColor.swift
```diff
@@ -0,0 +1,85 @@
...
+    case green = "green"
+    case purple = "purple"
+    case orange = "orange"
+    case pink = "pink"
+    case gray = "gray"
+    case brown = "brown"
+    case darkBlue = "dark_blue"
+    case clear = "clear"
+    
+    static func highlightColor(from string: String?) -> HighlightColor? {
```
nickmshelley: I think it would be better to take a non optional string and have the caller worry about unwrapping the optional.

nickmshelley:
👍  once tests pass again.

stephanheilner:
Tests pass now.

PR #74, opened 2017-03-19
Fix Travis configuration

hiltonc:


hiltonc:
This changes the defaults from `beta.lds.org` to `lds.org` and updates the client password in the test credentials for Travis CI. We should be able to rely on the Travis PR builds after this.

LDSAnnotationsTests/XCTestCase+LDSAnnotations.swift
```diff
@@ -43,8 +43,8 @@ extension XCTestCase {
             fatalError()
         }
         
-        let authenticationURL = UserDefaults.standard.string(forKey: "AuthenticationURL").flatMap { URL(string: $0) }
-        let domain = UserDefaults.standard.string(forKey: "Domain") ?? "beta.lds.org"
+        let authenticationURL = URL(string: UserDefaults.standard.string(forKey: "AuthenticationURL") ?? "https://www.lds.org/login.html")
```
stephanheilner: This might need to be `https://signin.lds.org/login.html`, but it it works with www, lets leave it as is.

hiltonc: It works, but I'd like to use the correct URL. Would `https://signin.lds.org/login.html` be more correct?

stephanheilner: Yes, that's the best url to use for sign in.

PR #75, opened 2017-03-22
Add test for security policy poopiness.

nickmshelley:


hiltonc:
Restarted the build, curious about the failure.

nickmshelley:
Oh I killed it because I knew it would fail.

hiltonc:
It fails beautifully :)

PR #77, opened 2017-03-31
Sync up/down as much as possible, even in the presence of errors.

nickmshelley:


nickmshelley:
I still have a bit of work to do on this (mainly anonymizing data) so don't merge yet. I just wanted to get it out there for maybe a preliminary review to make sure I didn't miss corner cases and that I have good test coverage.

LDSAnnotations/AnnotationStore+Annotation.swift
```diff
@@ -296,6 +296,15 @@ public extension AnnotationStore {
         }
     }
     
+    /// Returns annotations with uniqueIDs
+    public func annotationsWithIDsIn(_ IDs: [Int64]) -> [Annotation] {
```
hiltonc: What do you think about calling this `annotations(ids:)` or `annotations(idsIn:)`?

LDSAnnotations/AnnotationStore+SyncIssue.swift
```diff
@@ -0,0 +1,93 @@
...
+class SyncIssueTable {
+    static let table = Table("sync_issue")
+    static let id = Expression<Int64>("_id")
+    static let serviceID = Expression<String>("service_id")
+    static let message = Expression<String?>("message")
+    static let syncErrorType = Expression<SyncErrorType>("sync_error_type")
+    static let failureDate = Expression<Date>("failure_date")
+    
+    static func fromRow(_ row: Row) -> SyncIssue {
+        return SyncIssue(id: row[id],
```
hiltonc: Since when did you start manually splitting lines? 😉

LDSAnnotations/AnnotationStore+SyncIssue.swift
```diff
@@ -0,0 +1,93 @@
...
+                         syncErrorType: row.get(syncErrorType),
+                         failureDate: row[failureDate])
+    }
+}
+
+// MARK: Public
+
+public extension AnnotationStore {
+    /// Returns all sync issues ordered by failure date
+    public func allSyncIssues() -> [SyncIssue] {
```
hiltonc: You could just call this `syncIssues`, which would let you add defaulted filter arguments later. Alas, it would look like a method to sync things then 😢

LDSAnnotations/Error.swift
```diff
@@ -36,6 +30,28 @@ public struct SyncError: Error {
     let type: SyncErrorType
 }
 
+extension SyncError {
+    static func processJSONErrors(_ errorsJSON: [[String: Any]], defaultMessage: String, username: String, payloadJSON: Any) -> [SyncError] {
+        let errors: [SyncError] = errorsJSON.map { errorJSON in
+            let message = errorJSON["msg"] as? String ?? defaultMessage
+            let type: SyncErrorType
```
hiltonc: You can extract this into a private extension on SyncErrorType:

```swift
extension SyncErrorType {
    init(message: String) {
        switch message {
        case "Invalid label":
            self = .length
        case "Invalid tags":
            self = .invalidTag
        case "Invalid folders":
            self = .missingNotebook
        default:
            self = .unknown
        }
    }
}
```

LDSAnnotations/SyncAnnotationsOperation.swift
```diff
@@ -64,40 +59,29 @@ class SyncAnnotationsOperation: Procedure, ResultInjection {
...
-                    uploadLinkCount: self.uploadLinkCount,
+                let changes = SyncAnnotationsChanges(uploadServiceIDs: self.uploadServiceIDs,
                     downloadAnnotationCount: self.downloadAnnotationCount,
                     downloadNoteCount: self.downloadNoteCount,
                     downloadBookmarkCount: self.downloadBookmarkCount,
                     downloadHighlightCount: self.downloadHighlightCount,
                     downloadTagCount: self.downloadTagCount,
                     downloadLinkCount: self.downloadLinkCount)
-                completion(.success(notebooksResult: notebooksResult, localSyncAnnotationsDate: localSyncAnnotationsDate, serverSyncAnnotationsDate: serverSyncAnnotationsDate, changes: changes, deserializationErrors: self.deserializationErrors))
+                completion(.success(notebooksResult: notebooksResult, localSyncAnnotationsDate: localSyncAnnotationsDate, serverSyncAnnotationsDate: serverSyncAnnotationsDate, changes: changes, errors: self.nonfatalErrors))
```
hiltonc: Should we carry the name `nonfatalError` into the label in enum case `.success`?

LDSAnnotations/SyncAnnotationsOperation.swift
```diff
@@ -36,20 +36,15 @@ class SyncAnnotationsOperation: Procedure, ResultInjection {
...
+    var uploadServiceIDs = [String]()
     
     var downloadAnnotationCount = 0
     var downloadHighlightCount = 0
     var downloadBookmarkCount = 0
     var downloadNoteCount = 0
     var downloadLinkCount = 0
     var downloadTagCount = 0
-    var deserializationErrors = [Error]()
+    var nonfatalErrors = [Error]()
```
hiltonc: It seems odds to group this (via whitespace) with the download counters above it.

LDSAnnotations/SyncAnnotationsOperation.swift
```diff
@@ -123,43 +107,44 @@ class SyncAnnotationsOperation: Procedure, ResultInjection {
                 if let errorsJSON = syncAnnotationsIds?["errors"] as? [[String: Any]], !errorsJSON.isEmpty {
```
hiltonc: This block of code in response to a successful sync is pretty long. Maybe extract it to a private method?

LDSAnnotations/SyncIssue.swift
```diff
@@ -0,0 +1,59 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+import Foundation
+
+/// A Bookmark.
```
hiltonc: More a sync issue than a bookmark.

LDSAnnotations/SyncIssue.swift
```diff
@@ -0,0 +1,59 @@
...
+        self.id = id
+        self.serviceID = serviceID
+        self.message = message
+        self.syncErrorType = syncErrorType
+        self.failureDate = failureDate
+    }
+    
+}
+
+public func == (lhs: SyncIssue, rhs: SyncIssue) -> Bool {
```
hiltonc: These days I like to make an extension that conforms to `Equatable`, and then put the `==` implementation in that, like:

```swift
extension SyncIssue: Equatable {
    public static function == (lhs: SyncIssue, rhs: SyncIssue) -> Bool {
        ...
    }
}
```

LDSAnnotations/AnnotationStore+Annotation.swift
nickmshelley: I'd rather wait for the Swift compiler to catch up to the suggested naming convention. It's very obnoxious to me that this doesn't compile:

``` swift
func annotations(idsIn: [Int]) -> [Int] {
    return idsIn
}

let annotations = annotations(idsIn: [1, 2]) // Error: Variable used within its own initial value
```

But if you prefer I change it still I will.

LDSAnnotations/AnnotationStore+SyncIssue.swift
nickmshelley: Since the rest of the methods do and I match the style :).

(Although, to be honest, I did try it on one line first and actually found this more readable. Don't tell anyone!)

LDSAnnotations/AnnotationStore+SyncIssue.swift
nickmshelley: So does that mean you do or do not want it changed? :)

LDSAnnotations/SyncIssue.swift
nickmshelley: Heh good catch.

LDSAnnotations/AnnotationStore+Annotation.swift
hiltonc: Oh yeah, that's annoying. Nevermind.

LDSAnnotations/AnnotationStore+SyncIssue.swift
hiltonc: Our little secret.

LDSAnnotations/AnnotationStore+SyncIssue.swift
hiltonc: Meh :)

nickmshelley:
Ok I think this is ready to review/merge (once tests pass again).

stephanheilner:
I don't see anything that gives me heart burn. There are a lot of changes here though, but overall, I like what you're doing.

LDSAnnotations/SyncAnnotationsOperation.swift
nickmshelley: Oops, looks like I forgot this one. I changed _something_... must have been a different result.

digbads:
👍

PR #78, opened 2017-04-07
Handle the rest of the errors we're getting in GL.

nickmshelley:


LDSAnnotations/SyncErrorType.swift
```diff
@@ -25,21 +25,36 @@ import SQLite
 
 public enum SyncErrorType: Int64 {
     case unknown = 0
-    case length = 1
-    case deserialization = 2
-    case invalidTag = 3
-    case missingNotebook = 4
+    case deserialization = 1
```
hiltonc: I assume it's safe to renumber these?

nickmshelley: It is right now because we haven't shipped anything (even an alpha since that Jenkins job seems to be broken). But after we ship it won't be.

PR #84, opened 2017-05-08
Added fts search to notes

stephanheilner:


LDSAnnotationsTests/AnnotationStoreTests.swift
```diff
@@ -28,6 +28,20 @@ import XCTest
...
+        let docID = "12345"
+        
+        let annotation1 = try! annotationStore.addAnnotation(docID: docID, docVersion: 1, appSource: "Test", device: "iphone", source: .local)
+        _ = try! annotationStore.addNote(title: "I Am a Child of God", content: "And He has sent me here", annotationID: annotation1.id)
+        
+        let annotation2 = try! annotationStore.addAnnotation(docID: docID, docVersion: 1, appSource: "Test", device: "iphone", source: .local)
+        _ = try! annotationStore.addNote(title: "Book of Mormon Stories", content: "That my teacher tells to me", annotationID: annotation2.id)
+        
+        let results = annotationStore.searchNotes(searchString: "teacher")
+        XCTAssertFalse(results.isEmpty)
```
nickmshelley: It would probably be good to check that it's the right result, and that the snippet is what we expect (if that's not super hard to do). Can we also test searching for something in the title?

nickmshelley:
Some test suggestions. Also, it's not running travis tests because you did it from your fork. If you want travis to run, you need to make the branch in this repository (if I remember correctly).

LDSAnnotationsTests/AnnotationStoreTests.swift
stephanheilner: Yeah, on it.

stephanheilner:
Moved to a branch in LDSAnnotations so it will run tests: https://github.com/CrossWaterBridge/LDSAnnotations/pull/85

PR #89, opened 2017-05-15
Switch to new LDSAuthentication pod for auth

stephanheilner:


LDSAnnotations/SyncAnnotationsOperation.swift
```diff
@@ -22,6 +22,7 @@
 
 import Foundation
 import ProcedureKit
+import LDSAuthentication
```
SerenadeX: Seems meaningless to add this and nothing else to this file

LDSAnnotations/SyncNotebooksOperation.swift
```diff
@@ -22,6 +22,7 @@
 
 import Foundation
 import ProcedureKit
+import LDSAuthentication
```
SerenadeX: Seems meaningless to add this and nothing else to this file

LDSAnnotations/SyncAnnotationsOperation.swift
stephanheilner: It adds the AuthenticationOperation as a dependency operation. That class was moved from LDSAnnotations to LDSAuthentication which is why the only change that is needed is the import. But it uses the class that's in the LDSAuthentication framework now instead, so it needs it.

LDSAnnotations/SyncNotebooksOperation.swift
stephanheilner: Same as above.

LDSAnnotations/SyncAnnotationsOperation.swift
SerenadeX: Sounds good to me

PR #91, opened 2017-05-16
Add string description to error logger.

nickmshelley:


hiltonc:
Maybe not useful in all cases, but `#function` might be a useful macro for this.

stephanheilner:
I was trying to make the same comment! I think using #function is a better approach.

stephanheilner:
Just add a function like this:
```
static func logError(function: String = #function, line: Int = #line, error: Error) {
        let message = "\(function) (\(line))"
        logErrorHandler?(message, error)
    }
```
And use that instead of the static var.

nickmshelley:
Well, I wish I had talked to someone before wasting so much time on those stupid strings... 😅

PR #94, opened 2017-05-31
Improve logging.

nickmshelley:


hiltonc:
I recall reading somewhere that it is better to make the attempt and then use reachability to diagnose why it failed, rather than check reachability and then block the attempt.

nickmshelley:
This one would be good to squash and merge.

PR #95, opened 2017-06-15
Cache annotations to avoid so many lookups when syncing

stephanheilner:
?w=1 for a much cleaner diff


LDSAnnotations/Session.swift
```diff
@@ -140,6 +141,15 @@ public class Session: AuthenticatedSession {
             }
         }
     }
+    
+    public func urlSession(_ session: URLSession, task: URLSessionTask, didSendBodyData bytesSent: Int64, totalBytesSent: Int64, totalBytesExpectedToSend: Int64) {
+        let progress = totalBytesExpectedToSend > 0 ? Float(totalBytesSent) / Float(totalBytesExpectedToSend) : 0
+        progressObservers.notify(progress)
```
nickmshelley: Do we want to limit how often we do this? I'm not sure it will matter, but in other places it seems like we only do it every 5% or so to avoid performance problems.

LDSAnnotations/SyncAnnotationsOperation.swift
```diff
@@ -158,18 +161,27 @@ class SyncAnnotationsOperation: Procedure, InputProcedure {
         let previouslyErroredNotebookAnnotations = annotationStore.annotationsWithIDsIn(previouslyErroredNotebookAnnotationIDs)
         let modifiedAnnotations = (annotationStore.allAnnotations(lastModifiedAfter: after, lastModifiedOnOrBefore: onOrBefore) + previouslyErroredNotebookAnnotations).unique()
         
+        let annotationSyncCache = AnnotationSyncCache(annotations: modifiedAnnotations, annotationStore: annotationStore, fullTagsAndNotebooks: true)
+        
+        var index = 0
```
nickmshelley: I think you can get rid of this `var` by using `modifiedAnnotations.enumerated.map` below.

LDSAnnotations/SyncAnnotationsOperation.swift
```diff
@@ -193,13 +205,20 @@ class SyncAnnotationsOperation: Procedure, InputProcedure {
         let docIDs = syncIDs.flatMap { $0["docId"] as? String }
         let docIDsAndVersions = session.docVersionsForDocIDs?(docIDs)
         
+        var index = 0
```
nickmshelley: Same thing with `enumerated`.

LDSAnnotations/SyncAnnotationsOperation.swift
```diff
@@ -247,254 +266,234 @@ class SyncAnnotationsOperation: Procedure, InputProcedure {
...
-                    linksByAnnotationID[link.annotationID]?.append(link)
-                } else {
-                    linksByAnnotationID[link.annotationID] = [link]
-                }
-            }
-            
-            let notebookIDsByAnnotationID = annotationStore.notebookIDsWithAnnotationIDsIn(annotationIDs)
-            let tagsIDsByAnnotationID = annotationStore.tagIDsWithAnnotationIDsIn(annotationIDs)
             
+            var index = 0
```
nickmshelley: Use `enumerated`

LDSAnnotations/SyncAnnotationsOperation.swift
```diff
@@ -521,4 +520,57 @@ class SyncAnnotationsOperation: Procedure, InputProcedure {
...
     
+    func progressChanged(progress: Float) {
+        Log.debug("Progress: \(progress)")
+    }
+    
+}
+
+struct AnnotationSyncCache {
+    let annotationIDs: [Int64]
+    var noteByAnnotationID = [Int64: Note]()
```
nickmshelley: None of these should be modifiable outside of the init, so make them `let`s.

LDSAnnotations/AnnotationStore+Tag.swift
```diff
@@ -118,6 +118,30 @@ public extension AnnotationStore {
         }
     }
     
+    func tagsWithAnnotationIDsIn(_ annotationIDs: [Int64]) -> [Int64: [Tag]] {
```
nickmshelley: Remember to change this name if you decide to change the other.

nickmshelley:
.

nickmshelley:
👍

